PR #9, opened 2017-05-30
Fix minor HTML issues

daywaves:
Hi! I was just going through your series and noticed a few very minor issues with the HTML in the CSS/styling-text section. I figured I would fix them. Thanks for the excellent guide!


chrisdavidmills:
Yup, these look good - thanks!

PR #12, opened 2017-06-14
Moved <script> element into <body> element

hadleyel:


hadleyel:
Absolutely, no problem!

PR #13, opened 2017-06-15
Moved <script> into <body>, removed .focus()

hadleyel:
Moved <script> element into <body> element following W3C recommendations; removed inputForm.focus(), which caused the page at https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math to jump past the first few paragraphs of content to the console whenever newly accessed.


chrisdavidmills:
These changes look good - thanks for the work!

PR #15, opened 2017-06-18
Fix the [for] attribute of the password label

vhalbwachs:


chrisdavidmills:
Cool, thanks for the fix. I've also fixed the listing at 

https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/How_to_structure_an_HTML_form

PR #20, opened 2017-06-25
changed btn.onmouseout to btn.onclick

qdirks:
The article specifies to change btn.onclick to different values, but there is no btn.onclick. I believe it was meant to have btn.onmouseout be as btn.onclick. The user is supposed to change it to btn.onmouseout later in the experiment.


chrisdavidmills:
Thank you for spotting and fixing this! I think I had it set correctly in my original version, but then played with it and forgot to set it back to the correct code ;-)

PR #22, opened 2017-06-29
Update heroes.html

userGitHub34535:


chrisdavidmills:
This shouldn't be added to this file, as this is the starter file for the example. You can find the finished version in heroes-finished.html.

PR #25, opened 2017-07-15
Move <script> tag before the closing </body>.

psylone:
According to [the article](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML#Applying_CSS_and_JavaScript_to_HTML):

> The <script> element does not have to go in the head; in fact, often it is better to put it at the bottom of the document body (just _before_ the closing &lt;/body&gt; tag), to make sure that all the HTML content has been read by the browser before it tries to apply JavaScript to it (if JavaScript tries to access an element that doesn't yet exist, the browser will throw an error.)

Also, with the `<script>` after the closing `</body>` we'll not pass W3C validation:

<details>
  <summary>W3C validation result</summary>
  
</details><br>

Not sure, can we safely move all `<script>` tags before the closing `</body>` in all other cases in the repo.

### DIFF
--- a/html/introduction-to-html/the-html-head/css-and-js.html
+++ b/html/introduction-to-html/the-html-head/css-and-js.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>Meta examples</title>
-    
+
     <meta name="author" content="Chris Mills">
     <meta name="description" content="This is an example page to demonstrate usage of metadata on web pages.">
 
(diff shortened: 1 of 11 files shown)


chrisdavidmills:
Thanks for spotting this. Yes, the <script> elements should all be inside the </body>, not after it. Do you fancy updating your PR to fix all of these?

chrisdavidmills:
Well, all the ones you have found ;-)

psylone:
Sure!

psylone:
Found the rest cases with regexp, that should be fine now )

chrisdavidmills:
Awesome, thanks! I've also gone through and made any necessary changes to the articles.

PR #27, opened 2017-07-20
Update style.css

sunboy-y8:
nav element's class lost '#'


chrisdavidmills:
You are right - thanks for the fix!

PR #30, opened 2017-08-13
Update simple-else-if.html

Daidala2017:
adding temperatures  (nesting example)


chrisdavidmills:
This addition isn't really much use as is, because there is nothing that allows the user to set the temperature. I'd rather keep this example simple.

PR #31, opened 2017-08-21
Update index.html

fh408931:
The DOM input event is fired synchronously when the value of an <input>, <select>, or <textarea> element is changed. (For input elements with type=checkbox or type=radio, the input event does not fire when a user clicks the control, because the value attribute does not change.)


chrisdavidmills:
Yes, you are right - thanks for catching this!

PR #32, opened 2017-08-24
updated to have section tags

MiHawkStackOverFlow:
Have added section tags for:- 
1. Starting dates, 
2. Subjects of study  
3. Exotic dance moves


chrisdavidmills:
I don't want to add this to the example - it over complicates what we are trying to test here, and it would result in a number of other changes being needed on other pages too.

MiHawkStackOverFlow:
Ok. Understood. I hope the changes are not incorrect. Thanks for the reply !!!

PR #35, opened 2017-08-29
Update style.css

fh408931:
Class and pseudo-class of the specificity of 10,
Elements and pseudo-elements of the specificity of 1


chrisdavidmills:
Yes, you are right. Thanks for the fix!

PR #38, opened 2017-09-01
Update ajax.html

fh408931:
Request to return the response data is json string, can not use json object method, should be converted into json object


chrisdavidmills:
This example already works fine, as the data is already returned as a JSON object in line 29:

https://github.com/mdn/learning-area/blob/master/javascript/apis/introduction/ajax.html#L29

I do agree with your addition of the var keywords however. You could keep those if you get rid of the other changes.

PR #40, opened 2017-09-17
Update oojs-class-further-exercises.html

ricksimpsons:
eliminating a comma before "and" in case of 3 or more interests.


chrisdavidmills:
I think this is OK as is — on MDN we follow US English rules, which include the serial or oxford comma.

PR #41, opened 2017-09-26
time added to receiver-column class

vkWeb:
I think there's no need to add receiver-column class again to the time tag. That's all for now!
Your feedback will be appreciated.


chrisdavidmills:
You could structure the markup like this, but in a formal written letter the date is often put on a separate line, in its own paragraph. I prefer this structure, as I think the date the letter was sent is an important detail. So I am not making this change.

I also wouldn't mark someone down on an exam for structuring a sample letter like this, as it is personal opinion, rather than "right/wrong", unless of course the example explicitly asked for the date to be put in a separate paragraph.

vkWeb:
Okay. No problem with it @chrisdavidmills  :)

PR #42, opened 2017-09-27
Update: 'type' attribute removed

vkWeb:
Double quotes added instead of single quotes and 'type' attribute is removed as it is optional in HTML5. There are a lot of students learning from MDN so, we should make sure we follow latest and the best practices and encourage them to do so.

I'll be waiting for your feedback @chrisdavidmills 😃

### DIFF
--- a/html/introduction-to-html/structuring-a-page-of-content-start/index.html
+++ b/html/introduction-to-html/structuring-a-page-of-content-start/index.html
@@ -3,7 +3,8 @@
   <head>
     <meta charset="utf-8">
     <title>Birdwatching</title>
-    <link href='https://fonts.googleapis.com/css?family=Roboto+Condensed:300|Cinzel+Decorative:700' rel='stylesheet' type='text/css'>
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Normally, I'd totally agree. However, in this case the offending line is auto-generated by the Google Fonts service ;-)

So I am not sure whether it would be better to update it, or keep it as is - people experimenting with google fonts themselves might get confused because they see this stuff generated by Google Fonts and wonder why it is different from the stuff in our example.

vkWeb:
Earlier Google fonts used single quotes but now it is replaced with double quotes and 'type' attribute is removed as it is totally optional. 

You can see the below illustration:

Google Fonts
*Double quotes used and there's no 'type' attribute*

chrisdavidmills:
Ah, they changed it recently? OK, no worries then. Merging this now.

vkWeb:
Thanks for the merge. 🤗

PR #45, opened 2017-10-17
Create jogo-numero-erros.html

rubenfaria:
text translated to brazilian portuguese (pt-br)


chrisdavidmills:
Hi there! I'd prefer to keep translations of examples in separate repos (we already have a German translation started).

I've created a new one for pt-br examples here: https://github.com/mdn/learning-area-pt-br

Can you add the example there instead? Ideally use the same folder structure as the original

rubenfaria:
Thanks for the orientation, man!
I will do it!

PR #46, opened 2017-10-18
Update can-script.js

fh408931:
The first time after all the search is not satisfied, querySelector ('p') to find the element is the footer under the p element, added to the main element below, the second time after all search, p element has been removed , The page does not have a p element, and document.querySelector ('p') returns null


chrisdavidmills:
Cool, thanks for the fix!

PR #48, opened 2017-12-11
Adding style and css tags

RajuSuranagi:
This change is to support Active learning: applying CSS and JavaScript to a page.


RajuSuranagi:
Understood this is not needed.

PR #49, opened 2017-12-21
Update other-examples.html

jenniferlrx:
small fix
<input type="image"> do not acept value attributes.


chrisdavidmills:
Cool, thanks for the fix!

PR #50, opened 2017-12-26
Update style.css

DarkPurple141:
Strictly speaking if the html had an h2 + p unbound by a section it would still be indented. This is more to adjust for the spec.


chrisdavidmills:
ok, sounds good.

PR #52, opened 2018-01-31
Updated Article titles

ColemanTO:


chrisdavidmills:
Probably makes more sense; thanks!

PR #54, opened 2018-02-15
correct the spelling of temperatures

bshankar:


chrisdavidmills:
Thanks! I'm amazed no-one has pulled me up on that before ;-)

PR #55, opened 2018-02-18
Match output of unchecking in CSS editor

bripmccann:
The [instructions](https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Debugging_CSS#Active_learning_find_the_errors!) say to <q cite="https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Debugging_CSS#Active_learning_find_the_errors!">try clicking on the <code>&lt;h1&gt;</code> and unchecking the applied declarations to find out which one is causing the problem.</q> Unchecking an element in the CSS editor changes the element to a comment.

### DIFF
--- a/css/introduction-to-css/debugging-css-finished/style.css
+++ b/css/introduction-to-css/debugging-css-finished/style.css
@@ -35,7 +35,7 @@ header, footer {
 }
 
 h1, footer p {
-  margin: 0;
+  /*! margin: -70px; */
 }
 
 h1 {


chrisdavidmills:
Hrm, it doesn't do this for me. If you already have the devtools open, clicking on the &lt;h1&gt; in the DOM inspector will highlight it, and then show what CSS is applied to the &lt;h1&gt; in the CSS rules panel over to the right. Each declaration in there — when hovered over — will have a checkbox that you can check and uncheck to apply and unapply it. This is what I am on about.

Are you trying to do this in the starter file, or the finished version?

PR #58, opened 2018-02-20
what-is-js html updates

bripmccann:
Used HTML5's `<!DOCTYPE html>`, changed `lang` from `en` to `en-US`, added `<meta charset="utf-8">`, and added indentation.


chrisdavidmills:
Looks good, thanks!

PR #59, opened 2018-02-21
Remove redundant code from client-side storage examples

reddraggone9:
These also show up in code samples in [the article](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Client-side_storage). I didn't make the changes there since I'm not sure what the policy is for updating excerpts from a repo.

This probably isn't the place, but since I've got your attention, I found
> until after all the tasks inside the block have been completed.

to be oddly worded. It had me thinking some kind of magic was taking place based on a code block until I actually looked at the example and the related [documentation](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil) to see it was just waiting on the promise to settle.

Thanks for all of your work!

### DIFF
--- a/javascript/apis/client-side-storage/indexeddb/notes/index.js
+++ b/javascript/apis/client-side-storage/indexeddb/notes/index.js
@@ -126,9 +126,7 @@ window.onload = function() {
 
         // Set an event handler so that when the button is clicked, the deleteItem()
         // function is run
-        deleteBtn.onclick = function(e) {
-          deleteItem(e);
-        };
+        deleteBtn.onclick = deleteItem;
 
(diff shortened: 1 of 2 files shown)


chrisdavidmills:
Hey, thanks for helping to make the code better.

On the subject of the wording, I've changed it to "Inside the install handler we use the ExtendableEvent.waitUntil() method, available on the event object, to signal that the browser shouldn't complete installation of the service worker until after the promise inside it has fulfilled successfully."

Is that less confusing?

reddraggone9:
:+1: Yeah, that's good! Thanks again!

PR #62, opened 2018-03-22
Update main.js

redrambles:
Just a tiny change so that if the value for the customName field is female, there won't be a pronoun issue.


chrisdavidmills:
This is a really good point. Thanks for the addition!

PR #63, opened 2018-03-24
replace bgtile.png

2alin:
Original bgtile has an extra 1px border that makes the background not uniform, so I removed it. 
Before (look close to the bottom): 
screenshot from 2018-03-24 15-42-39
After:
screenshot from 2018-03-24 15-44-05


chrisdavidmills:
good call - thanks!

PR #65, opened 2018-03-26
can-store patches

2alin:
Two small fixes that I made:

1.  **Issue:** If you write something in the search box, then reload the page, what you wrote before remains (there's no problem there) but if you click the filter button without modifying the already existing search query, it won't filter anything and will show all the products.<br/>
 **Fix:** The problem happens because, after reloading, `lastSearch` takes the value of `searchTerm` (which is not empty), satisfies the condition checked in line 63 and stops the function before any filtering. The fix is easy, you just need to initialize `lastSearch` with an empty string; because when page loads, no search has been made yet.

2.  **Issue:** Trailing/leading whitespaces in search queries are taking into consideration, and I think they should be removed in case the user typed them by mistake.  <br/>
**Fix:** To trim the value of lastSearch when it's taking into consideration in the code.

### DIFF
--- a/javascript/apis/fetching-data/can-store/can-script.js
+++ b/javascript/apis/fetching-data/can-store/can-script.js
@@ -25,7 +25,8 @@ function initialize() {
 
   // keep a record of what the last category and search term entered were
   var lastCategory = category.value;
-  var lastSearch = searchTerm.value;
+  // no search has been made yet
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
This looks pretty good to me - thanks!

PR #66, opened 2018-03-27
nytimes patches

2alin:
Fixes I made:

1. **Issue:**  `current.lead_paragraph` returns 'undefined' and makes that every article shows no brief description of it; in fact such property is not listed in the object but it's still listed in the api documentation.<br/>
**Solution:**  `snippet` is property that gives a similar desired result, so I used that instead.

2. **Issue:** If we use queries like 'cats' or 'recipes' and leaving start-date/end-date empty, we end up with a  bunch of articles at the top of our search that are not real articles but 'topics' with broken links.<br/>
**Solution:** Filter search to items with 'article' as the value of  `document_type`, as explained [here](https://developer.nytimes.com/article_search_v2.json#/README#filtering-your-search).

3. **Issue:** `pageNumber` never resets which makes that every new search starts from the same page number that a previous search reached.<br/>
**Solution:** Implement a function similar to `nextPage`/`previousPage`, that will reset `pageNumber` every time a new search is made.

4. Well, this one was a very small code typo; there was `span.textContent += current.keywords[j].value + ' ';` but span is created at every `for` loop, so there's no previous value of it and there's no need of an addition assignment, a simple assignment will work exactly the same.

### DIFF
--- a/javascript/apis/third-party-apis/nytimes/index.html
+++ b/javascript/apis/third-party-apis/nytimes/index.html
@@ -72,10 +72,14 @@
       var displayNav = false;
 
       // Event listeners to control the functionality
-      searchForm.addEventListener('submit', fetchResults);
+      searchForm.addEventListener('submit', submitSearch);
       nextBtn.addEventListener('click', nextPage);
       previousBtn.addEventListener('click', previousPage);
-
+      
+      function submitSearch(e){
+        pageNumber = 0;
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
This looks really good, thanks!

2alin:
You're welcome and thanks to you, chris,  for the merge  :smiley:

PR #69, opened 2018-04-26
The paragraph tag missing

tkdphoenix:
The variable para points to a paragraph tag that doesn't exist.


chrisdavidmills:
You are right, but cans you put it beneath the div, like it is in the start file? Thanks.

tkdphoenix:
I realized after the fact that the paragraph tag was there.

PR #72, opened 2018-05-13
Update javascript-label.html

koraytugay:
When I choose cancel, the name that already exists should not be cleared.

### DIFF
--- a/javascript/introduction-to-js-1/what-is-js/javascript-label.html
+++ b/javascript/introduction-to-js-1/what-is-js/javascript-label.html
@@ -1,37 +1,39 @@
 <!DOCTYPE html>
 <html lang="en-US">
-  <head>
+<head>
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
I appreciate your feelings on this, but this is supposed to be a really simple trivial example to explain what JS allows you to do on top of HTML and CSS (see https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_is_JavaScript#A_high-level_definition).

And it does this quite well, in spite of the shortcoming you've pointed out. But if we did add in your code we would make the example more complicated, and I'm not sure that's such a good thing. At this stage the reader how not learnt what conditionals are

koraytugay:
@chrisdavidmills That is fine, thank you for your response.

PR #75, opened 2018-06-15
Missing parenthesis at the end of sentence

jzunigacoayla:
In table's caption from Planets data live example


chrisdavidmills:
Cool, thanks for the fix.

PR #78, opened 2018-07-02
Stray end tag for div element

jzunigacoayla:
Unintended div element in CSS layout positioning examples.


rachelandrew:
That's funny - I created a position: sticky one the other day and removed a stray div (as I saved the file from another one in the set) assuming that I had added it myself, not that they all had one. Thanks for the PR.

PR #79, opened 2018-07-08
Introduction to HTML

PegMat:
I changed some lowercase letters to uppercase letters in the title and <h1>. I also added unordered and ordered list tags.


PegMat:
okay, thanks

PR #81, opened 2018-07-12
update css-and-js.html

pr1smo:
I can't get this to open up the way that it is supposed to on my browser. I don't know what I'm doing wrong.

### DIFF
--- a/html/introduction-to-html/the-html-head/css-and-js.html
+++ b/html/introduction-to-html/the-html-head/css-and-js.html
@@ -2,8 +2,10 @@
 <html lang="en-US">
   <head>
     <meta charset="utf-8">
-    <title>Meta examples</title>
-
+<link rel="stylesheet" href="style.css">
+    
+<title>Meta examples</title>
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
What exactly is not working for you with this example? I've just tried it as it currently is in the repo, and it seems to still work fine.

pr1smo:
Either the CSS will work or the JavaScript will work but they don't work together. When the CSS is active I can't create a list but if the CSS isn't active I can. I don't know what I'm doing wrong. I've since moved on in the Mozilla course and all the other things I do work as they should.

chrisdavidmills:
OK, no problem then. I think that if other examples are working fine, then you should not worry about this. Probably some kind of subtle syntax error ;-)

chrisdavidmills:
By the way, if you have questions about something to do with an example in this repo, but you are not sure if something should be changed, you can file an issue instead of a pull request. See the "Issues" tab at the top of the repo.

PR #83, opened 2018-07-24
Remove the unreachable code

Ashugupta01:
Remove the last line of the unreachable code in the script part.

### DIFF
--- a/javascript/introduction-to-js-1/first-splash/number-guessing-game.html
+++ b/javascript/introduction-to-js-1/first-splash/number-guessing-game.html
@@ -116,8 +116,6 @@
         guessField.focus();
 
         lastResult.style.backgroundColor = 'white';
-
-        randomNumber = Math.floor(Math.random() * 100) + 1;
       }
     </script>
   </body>


chrisdavidmills:
But that line isn't unreachable. In fact, if you don't include that line, the number to guess is just the same on every single new game, which makes it a bit boring ;-)

Ashugupta01:
Oh I'm so sorry!
Fully ignored that fact. But I was kind of thinking of removing the duplicate lines from the code. Is there any other workaround for achieving what I was trying to do?
Thanks so much for reviewing and conveying the significance of that line. 👍

chrisdavidmills:
Hey, no worries.

In a real app, it would be better to make the code more DRY by defining a function that returns the random number and then calling that, rather than repeating the same code twice. But this is just a simple beginner's demo, and I think I was trying to cut down on the number of functions used to make it easier for beginners to understand.

We don't actually explain functions under later in the course.

Ashugupta01:
Oh okay, I see the bigger picture. Thanks again for the clarification Chris!

PR #84, opened 2018-07-25
Reference correct video source video file

rohit:
In the video and audio APIs tutorial the second source file references the `.mp4` file, but it should reference the `.webm` file.


chrisdavidmills:
You are totally right — thanks for the fix!

PR #86, opened 2018-07-28
fixing comparison operator in 3. EVENT LISTENER...

aka-yuki:
replacing != with the correct !== comparison operator inside if statement in
3. EVENT LISTENER AND PARTIAL FUNCTION DEFINITION


chrisdavidmills:
Good call; thanks for the update!

PR #91, opened 2018-08-25
My question

LittleMang:
I am confused, please compare the contents of these two URLs.
10 lines ：It was 94 fahrenheit 

[13 lines](https://github.com/mdn/learning-area/edit/master/javascript/introduction-to-js-1/assessment-start/raw-text.txt)：It was 94 farenheit
These two are different,I am using the Chinese tutorial at [MDN](https://developer.mozilla.org/zh-CN/docs/Learn/JavaScript/First_steps/Silly_story_generator#%E6%AD%A5%E9%AA%A4)
将 '94 farenheit' 置换成为变量 temperature 的值
According to the prompt I copied '94 farenheit',then an error occurred, I spent an hour finding out the problem.
This may cause problems for other learners.

### DIFF
--- a/javascript/introduction-to-js-1/assessment-finished/main.js
+++ b/javascript/introduction-to-js-1/assessment-finished/main.js
@@ -7,7 +7,7 @@ function randomValueFromArray(array){
   return array[random];
 }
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Ah, thanks for the fix. I've corrected it in the Chinese translation too.

PR #94, opened 2018-09-11
typo: fallback content wrong filename

torzsmokus:


chrisdavidmills:
Thanks for catching that! I've updated the code snippet inside the associated article too:

https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Other_embedding_technologies#The_%3Cembed%3E_and_%3Cobject%3E_elements

PR #96, opened 2018-09-18
Fix undefined variable

xuchunyang:


chrisdavidmills:
Yup, looks good. Thanks!

PR #99, opened 2018-09-28
Changed reference to female chickens from cock to hen.

irenesmith:
Cocks are male chickens, not female.


chrisdavidmills:
Yup, WFM!

PR #103, opened 2018-10-04
Update 'var' to 'let'

li3097:


javascript/building-blocks/loops/while.html
```diff
@@ -13,11 +13,11 @@
     
 
     <script>
-    var cats = ['Bill', 'Jeff', 'Pete', 'Biggles', 'Jasmin'];
-    var info = 'My cats are called ';
-    var para = document.querySelector('p');
+    let cats = ['Bill', 'Jeff', 'Pete', 'Biggles', 'Jasmin'];
```
codesandtags: I think that `cats` could be a `const` variable because it's content does not change.

chrisdavidmills: Yes, it could. I'm going to merge this PR for now, but feel free to open a new PR to change this.

And apologies to @li3097 for missing this PR for so long. I don't recall getting an email notification.

PR #104, opened 2018-10-04
Update 'var' to 'let'

li3097:


chrisdavidmills:
My apologies; I didn't get an email alert about any of your PRs, so had not seen these. They look generally good; One thing I would say is that we've started generally using const for references to DOM elements (because they are likely to be constant and not need to be updated). Can you update these?

Also, mention me by @name when done, just to make sure I don't miss the updates. Thanks.

PR #106, opened 2018-10-04
Update 'var' to 'let'

li3097:


chrisdavidmills:
Hi Lisa!

Thanks very much for your contributions. Unfortunately I can't merge these changes right now, and I have also reverted your changes to the [related article](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Looping_code). The reason is that - yes, I agree we should modernize this code - but we don't explain `let` (or `const`) earlier on in the course, therefore people coming to this section of the course and finding `let` are going to be really confused ;-)

So, we need a plan of action. Before we include your changes, we'd need to update the article that explains variables to add a description of let and const, what they are and why they are different to (and better than) var. This can be found here: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Variables

Of course, these folks are complete beginners, so we'd want to limit the complexity of explanation and defer some of it to later on (e.g. lexical scoping would confuse people at that point).

Once that work is done, I'd be happy to accept your changes.

So my question is — would you be happy to help me write this? ;-) Once that is written, further help with getting the course changed over to let and const would also be really appeciated.

Thanks again.

li3097:
@chrisdavidmills I will submit a new PR on the page you said.

li3097:
@chrisdavidmills I couldnt find the file where I thought the updating needed to be so the below is my suggestion. 

Declaring a variable
To use a variable you've first got to create it — more accurately, we call this declaring the variable. To do this, we type the keyword var followed by the name you want to call your variable:

var myName;
var myAge;

Here we're creating two variables called myName and myAge. Try typing these lines in now in your web browser's console, or in the below console (You can open this console in a separate tab or window if you'd prefer that). After that, try creating a variable (or two) with your own name choices.

The latest version of JavaScript adds more descriptive variable versions of 'var'. They are 'const' which means it is staying constant and 'let' which allows variables to change. Example would be const world = Earth; let year = 2018;

chrisdavidmills:
Hi there @li3097 — just to let you know that I've added some new sections to the variables page to explain the necessary bits about let and const at this stage:

* https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Variables#The_difference_between_var_and_let
* https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Variables#Constants_in_JavaScript

Let me know what you think!

I have also accepted your PR, reverted the loops article to publish your changes, and I will now go through and give it a check to see if you've missed anything.

Many thanks for your contriubutions!

li3097:
@chrisdavidmills I like it and you are a much better writer than me.

chrisdavidmills:
Cool, glad you liked it!

PR #108, opened 2018-11-02
Update complex-flexbox.html

stefsulzer:
Wrapping behavior of the buttons will now visually match the photograph finished example. 

Changed line 58 button's flex value to 1 auto. Equivalent to flex: auto but continues above explicit usage of flex-grow/flex-shrink.

### DIFF
--- a/css/css-layout/flexbox/complex-flexbox.html
+++ b/css/css-layout/flexbox/complex-flexbox.html
@@ -55,7 +55,7 @@
       }
 
       button {
-        flex: 1;
+        flex: 1 auto;
         margin: 5px;
         font-size: 18px;
         line-height: 1.5;


rachelandrew:
I think flex: 1 is correct here - which would set flex-basis to 0, giving equally sized items. Although I didn't build the example.

I think it would be a good idea not to change the text on the page until the PR has been discussed and merged though as now this is inconsistent.

chrisdavidmills:
I think I'd prefer to keep this as `flex: 1` — I don't think adding `auto` makes much difference in this case, and in addition it adds the need to explain what auto is doing in this context. I don't think it gains us much, and it adds some possibility of confusion.

chrisdavidmills:
Ah, wait. I just read your revision message on MDN about this making it work in Saf/Chrome. I'm doing a U turn; PR merged ;-)

Thanks for helping out with this.

PR #110, opened 2018-11-17
Update flexbox0.html

Abdulaziz-Sadi:
I have set section to display flex


chrisdavidmills:
This is the starter file for the flexbox exercises. It shouldn't have any flexbox CSS in it; what you've done here is the first step of the exercise _after_ downloading the source file.

So, closing.

PR #113, opened 2018-12-02
<html> element must have a lang attribute

alexphilin:
When configuring a screen reader, users select a default language. If the language of a webpage is not specified, the screen reader will assume it is the default language set by the user. This becomes an issue for users who speak multiple languages and access website in more than one language. It is important to specify a language and ensure that it is valid so website text is pronounced correctly.

For example, my VoiceOver is trying to read these links in Russian
[Screenreader testing](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility#Screenreader_testing) section
> Look at [good-semantics.html](http://mdn.github.io/learning-area/accessibility/html/good-semantics.html), and note how the headers are found by the screenreader and available to use for navigation. Now look at [bad-semantics.html](http://mdn.github.io/learning-area/accessibility/html/bad-semantics.html), and note how the screenreader gets none of this information. Imagine how annoying this would be when trying to navigate a really long page of text.

### DIFF
--- a/accessibility/aria/aria-div-buttons.html
+++ b/accessibility/aria/aria-div-buttons.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
     <title>ARIA div buttons</title>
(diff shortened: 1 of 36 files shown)


chrisdavidmills:
Thanks for this; this is an oversight we are intending to fix.

I'd prefer to settle on `en-US` as a standard for MDN language. Can you update this?

alexphilin:
Done.

chrisdavidmills:
Fantastic, thanks!

PR #115, opened 2018-12-26
Duplicate file in "introduction-to-js-1/what-is-js" folder

lfzyx:


chrisdavidmills:
Ah yes, don't think that should be there. Thanks!

PR #121, opened 2019-02-27
Update text-start.html

angelito-gm:


chrisdavidmills:
This is supposed to be the start point of the exercise, without any markup added to the content s the reader can try it themselves. So this isn't needed.

PR #128, opened 2019-04-14
Populate balls array outside the animation loop

campos20:
In the bouncing balls javascript example, the array to store object is populated inside the animation loop. This doesn't seem necessary since no ball is lost.

To keep things a bit cleaner, we can populate it outside the animation loop.

As far as I can tell, this would also require changes [here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_building_practice#Animating_the_ball) and a similar [here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Drawing_graphics#Animations).

### DIFF
--- a/javascript/oojs/bouncing-balls/main-finished.js
+++ b/javascript/oojs/bouncing-balls/main-finished.js
@@ -72,30 +72,30 @@ Ball.prototype.collisionDetect = function() {
   }
 };
 
-// define array to store balls
+// define array to store balls and populate it
 
 var balls = [];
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Hah, yeah, you're totally right here. This really doesnt need to be inside the loop. Thanks for the fix!

chrisdavidmills:
All related articles and other bits of code now updated too.

PR #130, opened 2019-04-18
add es6 version of main-finished.js

tamasmucs:
example ES6 version of the complete bouncing ball practice. 

In the object building practice article there is a guide to create a bounding balls animation demo. For educational reasons it is good to write ES5 style inheritance, but by now due to the wide ES6 support of browsers I think it makes more sense to at least link this file (if not replace the original one) in the tutorial.


chrisdavidmills:
I am happy to have the ES class version in the repo, but I'll have to have a think about what to do in terms of updating the tutorial. I think it is nearing the time to update this stuff to ES6 syntax, but I don't have the time right now.

PR #134, opened 2019-05-01
Update index.html to clean JavaScript code

escape0707:
I tested and think this approach could be clearer and easier to understand.
If it's valid, I'll then go and change corresponding documentation page.

### DIFF
--- a/html/forms/indeterminate-example/index.html
+++ b/html/forms/indeterminate-example/index.html
@@ -43,22 +43,22 @@
   }
 
   function updateDisplay() {
-    var checkedCount = 1;
+    var checkedCount = 0;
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
I've tested this, and yes, it works as expected, and seems a bit more intuitive.

Thanks for the contribution! And thanks also for updating the docs page.

escape0707:
Thanks for reviewing and also for maintaining such helpful documentations actively! They helped me a lot!
[Documentation page updated.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox$compare?locale=en-US&to=1547750&from=1544591)

chrisdavidmills:
Perfect, thanks!

PR #135, opened 2019-05-02
Update basic-for-improved.html

mgo7:
I think there should be  if(i === cats.length -1 ) - otherwise it will never run this code?

### DIFF
--- a/javascript/building-blocks/loops/basic-for-improved.html
+++ b/javascript/building-blocks/loops/basic-for-improved.html
@@ -18,7 +18,7 @@
     const para = document.querySelector('p');
 
     for(let i = 0; i < cats.length; i++) {
-      if(i === cats.length - 1 ) {
+      if(i === cats.length - 1) {
         info += 'and ' + cats[i] + '.';
       } else {
         info += cats[i] + ', ';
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
You are right, this doesn't work! It is funny, this code is correct in the article, but wrong in the GitHub repo.

Thank you for pointing it out, and providing a fix.

I did have a nitpick about it — the code is correct, but your spacing isn't quite right. Can you change the spacing to 

```
if(i === cats.length - 1)
```

then I'll merge it.

Thanks again!

javascript/building-blocks/loops/basic-for-improved.html
```diff
@@ -18,7 +18,7 @@
     const para = document.querySelector('p');
 
     for(let i = 0; i < cats.length; i++) {
-      if(i === cats.length ) {
+      if(i === cats.length - 1 ) {
```
escape0707:
```suggestion
      if(i === cats.length - 1) {
```

chrisdavidmills: Ummm, yup? ;-)

chrisdavidmills:
Perfect, thanks for that. Merging.

PR #137, opened 2019-05-11
update can-style.css

Kowalatam:
Change html background color from yellow to gray-blue. It was too bright for my eyes :)

image

image


chrisdavidmills:
I appreciate your enthusiasm, and to be fair your choice is probably better than mine (I am color blind too, so I've got no hope). But I'd prefer not to accept this change — we'd also have to change this in the other versions of this example, change the icon and heading colors in the example to match the background, and update all the screenshots in the associated tutorial.

Thanks anyway ;-)

PR #139, opened 2019-05-24
Input Box Case sentivity

mayunited12:
Input box will now be case insensitive hence, can accept any valid contact name in any case.


chrisdavidmills:
This matches what we've got in the corresponding article (https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Looping_code) too. Did you change it there too?

In any case, thanks for the addition. I think this is useful.

PR #146, opened 2019-07-25
Updating var to let.

kasai-senshi:


chrisdavidmills:
Thanks for helping to modernize the code!

I would like to request a change here however — it is best practice to use `const` for variables whose values never change after they are declared, and `let` for variable whose values do change during the execution of the code.

avivmu:
@kasai-senshi please either update or close

PR #150, opened 2019-08-27
Update letter-text.html

EricHampton12:


chrisdavidmills:
This is the starting text for the exercise, so doesn't need to be marked up

PR #151, opened 2019-08-29
Dealing with reset button's issue.

jasontwuk:
Letting users to use the reset button straight after they press the start button without the clock having wrong behavior. Originally, if a user starts the clock and doesn't press the stop button before pressing the reset button, the clock will reset to 00:00:00 and then start running again automatically.


chrisdavidmills:
Yeah, I agree with you that this behavior is better. Thanks for your contribution!

jasontwuk:
Cheers :)

PR #154, opened 2019-08-30
Update reaction-game.html

jasontwuk:
(1) Match up the code with the previous example's (simple-raf-spinner.html). 
(2) Remove a let declaration, because it is repeated. 
(3) The contents in this MDN web docs webpage (https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Timeouts_and_intervals#requestAnimationFrame()) might need an update as well. The location is "Active learning: a reaction game - 4" in that webpage.


chrisdavidmills:
I've updated the article too.

PR #156, opened 2019-09-13
Update dom-example.html

sebastian-sugarai:
feat/add-features-to-dom-example #999


chrisdavidmills:
This is not needed. The file as it exists is the starting state for the exercise, and the instructions in the page tell you to add the `<script>` element.

PR #159, opened 2019-10-13
Use ES6 variable keywords

hararca:
Window's height and width have to be stored in as a "let" to be able to change their values on resize. Otherwise the program throws an error.

### DIFF
--- a/javascript/apis/document-manipulation/window-resize-example-finished.html
+++ b/javascript/apis/document-manipulation/window-resize-example-finished.html
@@ -24,18 +24,18 @@
     </div>
 
     <script>
-      var div = document.querySelector('div');
-      var WIDTH = window.innerWidth;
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
I've tried this in multiple browsers, and I don't get an error if var is used. Which browser are you trying this in?

hararca:
Sorry for the misconception.
What I really meant is that if we want to modernize the syntax and make use of const and let, then we cannot treat WIDTH and HEIGHT as const since the value have to be updated afterwards.
There is no problem with the good 'ol var! :)

chrisdavidmills:
Ah, I see what you mean now! Yup, I am happy with this change.

And it looks like the changes have already bene made in the associted article:

https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Manipulating_documents

PR #162, opened 2019-10-15
Simplify the formula to get the minutes

hararca:
I think this is easier to understand


chrisdavidmills:
But it's not quite as accurate. I'd rather add extra explanation to the article than simplify the formula.

hararca:
OK, you're right. Now I see the point.

PR #163, opened 2019-11-01
Using srcObject instead of createObjectURL

wwags33:
createObjectURL no longer accepts MediaStreams, so the previous version threw a TypeError.


chrisdavidmills:
Cool, thanks for the fix!

PR #166, opened 2019-11-14
Update the info box color and margins to match the capture.

snakada:
@chrisdavidmills 

It seems that the capture has been replaced in https://github.com/mdn/sprints/issues/2307, so I tried adjusting the example so that it looks closer.

Can you adopt this?


chrisdavidmills:
Super, thanks so much for doing this! I was intending to do it today ;-)

I have also updated the fixed-info-box exampe to use the same colors/styling.

PR #167, opened 2019-11-21
Fallback example fix for Internet Explorer

thinw002:
The intention of this example seems to be to illustrate how the fallback is enabled in browsers that don't support input type="time". The statement test.type = 'time'; fails in Internet Explorer 11, the obvious example of a browser that doesn't support type="time". I've changed the example so that it works in the Internet Explorer.

### DIFF
--- a/html/forms/datetime-local-picker-fallback/index.html
+++ b/html/forms/datetime-local-picker-fallback/index.html
@@ -92,7 +92,13 @@ fallbackLabel.style.display = 'none';
 
 // test whether a new date input falls back to a text input or not
 var test = document.createElement('input');
-test.type = 'date';
+
+try {
+  test.type = 'time';
+} catch (e) {
+  console.log(e.description);
+}
+    
 // if it does, run the code inside the if() {} block
 if(test.type === 'text') {
   // hide the native picker and show the fallback


chrisdavidmills:
Hi there @thinw002, and thanks for pointing this out. You are indeed right tha my test wasn't working properly.

I tried out your fix and found that it wasn't working quite as expected either, but then I realised that `time` should be `datetime-local` for this one.

However, by that point I'd uploaded the fix to GitHub by myself, so this PR is no longer needed - sorry about that! I really appreciate you finding the fix though.

Saying that, all of the other similar examples need fixing, and the examples embedded on the pages need fixing too:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#Examples
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#Examples
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/month#Examples
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Examples
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/week#Examples

Would you be willing to update the other examples and their corresponding MDN pages? I'd really appreciate the help there too.

PR #170, opened 2019-12-09
gender neutral / tel for credit card

estelle:
Removed Miss and Mister and made it ace, king, queen to make it less gendered.
Changed credit card to tel instead of number, since a spinner is not correct for a cc#

### DIFF
--- a/html/forms/html-form-structure/payment-form.html
+++ b/html/forms/html-form-structure/payment-form.html
@@ -17,15 +17,21 @@
               <legend>Title</legend>
               <ul>
                   <li>
+                    <label for="title_2">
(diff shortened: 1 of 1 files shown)


estelle:
Yes, tel is used to enable getting a numeric dynamic keyboard without the number spinner. This is mentioned later in the html5 input types article.

estelle:
Oh, and it's also in the text above the example code in the tutorial. https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/How_to_structure_an_HTML_form "the second is an <input> element of type tel, for entering a credit card number (while we could have used the number type, we don't want the number's spinner UI). "

chrisdavidmills:
OK, wfm!

PR #172, opened 2019-12-13
Update raw-text.txt

Coach-G:
the MDN exercise file is missing "random" variable


chrisdavidmills:
Well, the two versions do the same thing, but I think it should be consistent, and the version you are updating this to is easier to understand for beginners.

Thanks for the contribution!

PR #174, opened 2019-12-24
changed from absolutes positioning to grid

estelle:
We should not be suggesting absolute positioning as a potential solution, so switched it to grid.


chrisdavidmills:
LGTM!

PR #176, opened 2019-12-28
Uppercase "squadName"

ckashby:
Looks better.


chrisdavidmills:
Yup, probably does ;-)

PR #178, opened 2020-01-02
changed alignment of cells

Lisagrant101:
The issue said some things were not lining up (space to the left of some words)
This was because text align was center and some words were shorter so they were offset by center
I added th to the td selector in the css file and aligned all th and td cells to left


otisgbangba:
Many thanks Lisagrant for your input. I appreciate your assistance.

PR #181, opened 2020-01-24
removing global variable products

gitmrwick:
removing global variable in favour of passing variables between functions


chrisdavidmills:
LGTM; thanks @gitmrwick !

PR #182, opened 2020-01-24
Swapped input and label order

ughurh:
for consistency with the MDN web docs

https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_an_HTML_form#Labels_are_clickable_too!


chrisdavidmills:
Thanks for the fix!

PR #183, opened 2020-01-26
Fix duplicate id attribute in form input/label fields

83C10:
id="title_2" was repeated twice


chrisdavidmills:
@83C10 Great, thanks for the fix!

PR #184, opened 2020-02-02
Loop 1

abdeldjalilhachimi:
I have just tried to solve this task I hope I make it up


chrisdavidmills:
The code looks correct, but I don't really want it added to the starter file for the example ;-)

In future, to get your work assessed, follow the instructions at https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Test_your_skills:_Loops#Assessment_or_further_help

thanks!

PR #186, opened 2020-02-07
Use same text as in marking-up-a-letter-start file

jrsnzc:
accomodate > accommodate
livestock or be "fruitful" > livestock to be fruitful
imperceptably > imperceptibly
excellent > awesome


chrisdavidmills:
@jrsnzc thank you for the fixes!

PR #188, opened 2020-02-23
The To Upper Case methos was missing

praszad:
The To Upper Case was handled behind the screen and it was not in the code.


chrisdavidmills:
The uppercasing is being done via CSS in this case — `text-transform: uppercase;`

PR #189, opened 2020-02-23
Null || Undefined issue if No Name Given

praszad:
getting Null or undefined if the user fails to enter any char on the prompt


chrisdavidmills:
Your change doesn't really solve the problem. You still get a potentially confusing result if you don't enter a name, or click cancel.

But in this case, that's OK — this is supposed to be a really simple example to demonstrate generally what JavaScript can do for you, on top of HTML and CSS ([see here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_is_JavaScript#A_high-level_definition) for context). I don't want it to be any more complex than it has to be.

So respectfully closing this PR.

PR #191, opened 2020-03-11
Update links1-download.html

ragu0713:
here by i have included my code...if any changes in my code let me know...


chrisdavidmills:
Your changes look like they make sense, but I don't want them added onto my starter file. The starter file shouldn't contain the answer ;-)

PR #199, opened 2020-03-25
.innerHTML hint

tomigm:
When writing a line that contain "list", it transforms it to a String instead of a Node, so when .appendChild appears it throws this error in console:

**// Error: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'** 

My solution was:
 
   ```
 // Add your code here

let i = 0

while (i < myArray.length) {
    list.innerHTML += "<li>" + myArray[i] + "</li>";

    i++;
    
}

list.innerHTML += "</ul>";

    // Don't edit the code below here!
```

### DIFF
--- a/javascript/building-blocks/tasks/loops/loops1-download.html
+++ b/javascript/building-blocks/tasks/loops/loops1-download.html
@@ -30,6 +30,8 @@
     let list = document.createElement('ul');
 
     // Add your code here
+    
+    // list strings should include > list.innerHTML
 
     // Don't edit the code below here!
 


chrisdavidmills:
This isn't helpful at all really, plus you shouldn't even be using`innerHTML` - it's seen a bad practice these days. You should really be creating a new `<li>` instance on each loop iteration, then adding the array item at its textContent, before appending the `<li>` as a child of `<ul>`.

The https://wiki.developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Test_your_skills:_Loops#DOM_manipulation_considered_useful section on the page hints that you need to use such code, and points to an article to explain how to do it.

Also, you don't need `list.innerHTML += "</ul>";` - the `list` variable contains a `<ul>` element, which already has its own closing tag.

deeptipant:
IS there anyone to help me understand from scratch? Online personal one- on- one tutoring ? Please. Connect Or Skype: doma.deepti

PR #201, opened 2020-04-02
Make catch() called if URL is misspelled, as mentionned in tutorial

VincentDaviet:
Fix to call catch() method when a HTTP error occurs. 
For example when the URL trying to be fetched has been misspelling, as mentionned on the tutorial here : 
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Promises#Responding_to_failure

Otherwise the catch() method is never called, and the tutorial does not work at this point.


chrisdavidmills:
Thanks @VincentDaviet , this is useful!

VincentDaviet:
Thanks for that, that's my first real PR. What a pleasure to see it has been merged to master ! Fiesta!! :)

PR #203, opened 2020-04-04
Expand list of valid image types

apoorva-sriv:


chrisdavidmills:
Seems reasonable; thanks!

PR #204, opened 2020-04-05
Change variable ‘spinner’ value to ‘true’

hyfyooie:
The pause and continue action of the icon won’t work, if the variable is set to ‘false’. Also, added ‘draw()’ at the bottom to initialise the function.


chrisdavidmills:
As far as I can see, this isn't needed. The variable starts at `false`, and then the following event handler code handles running `draw()` if the variable is set to false, then setting the variable to true. Then if it is set to `true` when the body is clicked, it stops the animations by running `cancelAnimationFrame()`, and toggles the variable back to `false`, ready for the next click. 

```
document.body.addEventListener('click', () => {
  if(spinning) {
    cancelAnimationFrame(rAF);
    spinning = false;
  } else {
    draw();
    spinning = true;
  }
});

```

PR #207, opened 2020-04-10
Correct Chemical Formula for Water

JTansey93:
Water has two hydrogen atoms and one oxygen atom, not 20 hydrogen atoms! :)


chrisdavidmills:
LOL, good spot ;-)

PR #209, opened 2020-04-17
Refactor IF condition to make it more simple

VincentDaviet:
It seems to me that we do not need to use a "else" condition and force a return.


chrisdavidmills:
Looks good - thanks!

PR #210, opened 2020-04-20
Revert "Refactor IF condition to make it more simple"

chrisdavidmills:
Reverts mdn/learning-area#209


avivmu:
I don't understand what was the motivation for the original refactor.
Anyway maybe add why it needs to be reverted. 
Also are you waiting for something in order to merge it?

PR #213, opened 2020-04-30
Fix bug: Incorrect database name used

hyfyooie:
It was targeting the wrong database name, through and through.


chrisdavidmills:
The database is called `notes_db`. The object store inside the database is called `notes_os`. So this works fine as is.

PR #217, opened 2020-05-07
Bring bouncing balls JS in line with tutorial text

nfelger:
On https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_building_practice, the random function generates a number in [min, max], whereas in the provided main.js file it is [min, max).


chrisdavidmills:
I'm not sure what your comment mean here, but I can see that your change is useful in giving this code parity with what the tutorial says. Thanks!

PR #218, opened 2020-05-07
Copy update from main.js to other main-*.js files

nfelger:
I just noticed the difference from #217 is also present in two other files, so I changed those, too. Hope this is useful!


chrisdavidmills:
That's great; thanks!

PR #221, opened 2020-05-21
Update 6_fixed-positioning.html

haond10adp:
centering an element using `margin: 0 auto;` doesn't work without the explicit `left: 0` and `right: 0` combination (in case we want to center the element within the full viewport width)


chrisdavidmills:
I agree that the center effect doesn't work when the element is fixed positioned, but removing it altogether breaks the example, as there is then a gap between the top of the heading and the top of the screen.

Instead, cna you change the declaration to `margin-top: 0;` ?

chrisdavidmills:
Perfect, thanks!

PR #222, opened 2020-05-29
events2 exercise: Without tabIndex the canvas can't get focus

KjellKod:
Without tabIndex the canvas can't get focus and the exercise won't work, without adding exactly this within the HTML area which is  above `// add your code here`

### DIFF
--- a/javascript/building-blocks/tasks/events/events2-download.html
+++ b/javascript/building-blocks/tasks/events/events2-download.html
@@ -25,7 +25,7 @@
     <section class="preview">
     </section>
 
-    <canvas width="480" height="320">
+    <canvas width="480" height="320" tabindex="0">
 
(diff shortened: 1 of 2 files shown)


chrisdavidmills:
@KjellKod You can complete this example without needing to focus the canvas if you use a `key*` event on the `window` object rather than on the `<canvas>` element, but I think this addition is still a good one — it increases the reader's options for how they can complete the assessment, and makes keyboard a11y somewhat better.

Can you also make this change to the `<canvas>` element in the events2.html file too?

Many thanks.

KjellKod:
Sure thing

KjellKod:
@chrisdavidmills the PR is updated

chrisdavidmills:
Perfect; thanks!

PR #223, opened 2020-06-03
Put script inside head and use defer attribute

siraj-samsudeen:
The tutorial recommends this as the best way - hence, the code should use that method rather than using <script> before the closing body tag.


chrisdavidmills:
Yes, I agree - thank you for the update!

siraj-samsudeen:
Thank you Chris for your quick and prompt actions. Much appreciated.
Kind Regards,
Siraj, Chennai.
+91 97 899 68 576

PR #229, opened 2020-06-08
Add language to html tag and encode "|" in the URL"

avivmu:
Using [W3c Nu HTML validator](https://validator.w3.org/nu/) The following errors appears:

1. Warning: Consider adding a lang attribute to the html start tag to declare the language of this document.
2. Error: Bad value https://fonts.googleapis.com/css?family=Roboto+Condensed:300|Cinzel+Decorative:700 for attribute href on element link: Illegal character in query: | is not allowed.
This PR fixes those.


chrisdavidmills:
Looks good, thanks!

PR #230, opened 2020-06-15
Update punk-bands-unstyled.html

vgiffin:
Updated track title on "If The Kids Are United" to match the conventional capitalization, which is used in the punk-bands-complete.html file.

### DIFF
--- a/css/styling-boxes/styling-tables/punk-bands-complete.html
+++ b/css/styling-boxes/styling-tables/punk-bands-complete.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
     <title>UK punk bands</title>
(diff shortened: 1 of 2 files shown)


chrisdavidmills:
@avivmu agreed this would also be good.

@vgiffin can you also add lang="en" to the opening `<html>` tag? I must admit I was really bad at doing this when I first created these examples. Thanks!

avivmu:
Thanks. one thing only.Let's make [git commit history ](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)look nicer. You've done 2 changes, so I expect either one or two commits, no more. Also try to improve the [commit message](https://www.google.com/search?q=commit+message+guidelines&oq=commit+&aqs=chrome.1.69i59l2j69i57j0l2j69i60l3.2191j0j7&sourceid=chrome&ie=UTF-8)

vgiffin:
@avivmu I made all previous edit in the GitHub web UI, cloned locally to try to fix the commit message (and squash commits), but I'm not sure this is better.

chrisdavidmills:
@vgiffin this works fine for me; thanks for the improvements!

PR #231, opened 2020-06-19
End the game on "a" or "l" only

mrmowji:
Based on the description on MDN, pressing a wrong key shouldn't end the game, right?

> Regardless of which one of the player control keys was pressed, remove the `keydown` event listener using `removeEventListener()` so that once the winning press has happened, no more keyboard input is possible to mess up the final game result.


chrisdavidmills:
Ha, I never thoughts of this ;-)

Good fix, thanks!

PR #240, opened 2020-07-13
Update simple-raf-spinner.html

DrGaud:
Updated the CSS with 'position:absolute;' this was to keep the spinner rotating on the spot instead of rotating around the screen.


chrisdavidmills:
@DrGaud I've had a play with this, and I'm not sure it really makes much of a difference?

I think what we really need to do here is find a properly round spinner image, and set that as a background image on the div, make sure the div is a square, and rotate that.

KooXme (approved):

PR #243, opened 2020-07-18
removing redundant lines

Sonntag09:
removed task1 lines from task2 file to avoid confusion


chrisdavidmills:
Good fix @Sonntag09 , thank you!

PR #244, opened 2020-07-18
Added a missing element name

avivmu:
[Task](https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Typesetting_a_homepage#Project_brief)


chrisdavidmills:
Looks good, thanks @avivmu !

PR #246, opened 2020-07-19
Line 45: story's textContent to be empty string

Saif-Shines:
The story paragraph's initial value is intends to be empty string. Seems like that part is missed in here.


Saif-Shines:
Makes sense. Student when copy and pastes his/her code would intuitively expect to be syntactically correct. Yet, ignore the PR if this was a conscious choice.  😃

chrisdavidmills:
Yup, this was a conscious choice, so closing.

But thanks for the interest here @Saif-Shines! It's always better to question than to blindly accept things that don't look right on first glance.

PR #250, opened 2020-07-21
Update conditionals2-download.html

sadagatasgarov:


chrisdavidmills:
This is the starter file; I don't want the solution to be included in it.

PR #252, opened 2020-07-23
Update dom-example.html

ksmmaloney:


chrisdavidmills:
This is the starter file, so doesn't want the final code included in it.

PR #276, opened 2020-08-20
Update marking.md

RafaelGreen1:
fix the email validation regex
A-z is not correct because it will catch also these characters: [,\\,],^,_` (ascii 91-96)
and . should be replaced by \\.


chrisdavidmills:
Looks good; thanks @RafaelGreen1

PR #279, opened 2020-08-22
quicker loop!

misterybodon:
if we invert the conditions, the loop will run quicker

(sorry, I will next propose all changes in one commit, instead of so many pull requests!)

### DIFF
--- a/javascript/oojs/advanced/oojs-class-inheritance-start.html
+++ b/javascript/oojs/advanced/oojs-class-inheritance-start.html
@@ -68,10 +68,10 @@
           // all, adding each one to the main string followed by a comma,
           // except for the last one, which needs an and & a full stop
           for(let i = 0; i < this.interests.length; i++) {
-            if(i === this.interests.length - 1) {
-              string += 'and ' + this.interests[i] + '.';
-            } else {
+            if(i !== this.interests.length - 1) {
               string += this.interests[i] + ', ';
+            } else {
+              string += 'and ' + this.interests[i] + '.';
             }
           }
         }


RafaelGreen1:
if you really want to make the loop run quicker, just move the condition out of the loop
and handle the last element separately.
by the way, the topic of the article has nothing to do with time performance optimization,
so I'm not sure if it worths changing it at all

misterybodon:
I don't care about your opinion. If you really want someone to care, just be friendly. 
Did you understand plastic socket?

chrisdavidmills:
@misterybodon @RafaelGreen1 thank you for the discussion here folks. Yes, the article doesn't have anything to do with performance optimization, but it is worth considering making an example better if it:

1. Makes the code easier to understand for the target audience.
2. Makes the code better practice.
3. Removed something that is bad practice.

I am all for presenting implicit best practice code, even if we don't discuss all the aspects of why it is best practice right away (which might confuse the reader and not help at this stage).

So, considering these points, what do we think is best to do here? I am happy to take a look at this later in the week, once I find the time. But your help is always appreciated.

RafaelGreen1:
in my opinion the code is good enough as it is now
it's easy to understand and pretty good
as I mentioned before, it may be slightly improved by moving the condition out of the loop
as it make the code shorted, faster and easier to understand (just in my opinion of course)
```
let i;
for(i = 0; i < this.interests.length - 1; i++) {
  string += this.interests[i] + ', ';
}
string += 'and ' + this.interests[i] + '.';
```

BassOfBass:
Putting the cycle counter outside of the loop condition is kinda meh, considering the keyword "let" was created exactly to counter the problem of "var" leaking out of loop scopes, amongst the other things.

RafaelGreen1:
@BassOfBass I agree.
What about this one?
```
for(let i = 0; i < this.interests.length - 1; i++) {
  string += this.interests[i] + ', ';
}
string += 'and ' + this.interests[this.interests.length - 1] + '.';
```

BassOfBass:
@RafaelGreen1 
Why bother with explicit loops with string concatenations when you can operate on an array until Array.join() at the very end?
Something like this:
```
function checkInterests(interests) {
  if (interests.length === 1) return interests[0];

  const temp = Array.from(interests);
  const lastInterest = temp.pop();
  const beforeLast = temp.join(", ");
  return beforeLast + " and " + lastInterest;
}
```

RafaelGreen1:
@BassOfBass 
in this case we know that ``` interests.length > 2 ```
so we can use this:
```
string += `${interests.slice(0, -1).join(', ')} and ${interests[interests.length - 1]}`;
```

avivmu:
@chrisdavidmills I suggest closing this since the focus of the article is about Object Oriented JS and the suggestions here don't improve the readability of the code (and performance is hardly affected, if at all)

PR #280, opened 2020-08-24
Update validation.js - fix firefox focus problem

RafaelGreen1:
return false cause the browser to not opening the link which fixes the problem
need to update the documentation as well

### DIFF
--- a/accessibility/css/validation.js
+++ b/accessibility/css/validation.js
@@ -40,6 +40,7 @@ function createLink(testItem) {
   anchor.href = '#' + testItem.input.name;
   anchor.onclick = function() {
     testItem.input.focus();
+    return false;
   };
   listItem.appendChild(anchor);
   errorList.appendChild(listItem);


avivmu:
Any context? What is the relevant documentation?

RafaelGreen1:
> Any context? What is the relevant documentation?

in this article: https://developer.mozilla.org/en-US/docs/Learn/Accessibility/CSS_and_JavaScript

Note: The focus() part of this example is a bit tricky. Chrome and Edge (and newer versions of IE) will focus the element when the link is clicked, without needing the onclick/focus() block. Safari will only highlight the form element with the link on its own, so needs the onclick/focus() block to actually focus it. Firefox doesn't focus the inputs properly at all in this context, so Firefox users can't take advantage of this at present (although everything else works fine). The Firefox issue should be fixed soon — work is being done to give Firefox behaviour parity with other browsers (see bug 277178).

Thanks for looking at it

accessibility/css/validation.js
```diff
@@ -40,6 +40,7 @@ function createLink(testItem) {
   anchor.href = '#' + testItem.input.name;
   anchor.onclick = function() {
     testItem.input.focus();
+    return false;
```
avivmu, line 43: 1. According to [this SO question](https://stackoverflow.com/questions/128923/whats-the-effect-of-adding-return-false-to-a-click-event-listener) one should use `event.preventDefault()` instead of returning `false`.
2. Since this is a workaround until the [mentioned bug](https://bugzilla.mozilla.org/show_bug.cgi?id=277178) would be fixed, I would add a comment to remove it once the bug is fixed

RafaelGreen1: is the mentioned bug open for 16 years or am I imagining?

avivmu: Actually I wouldn't bother changing the code at all (it's a workaround).
You might want to edit the documentation to mention that adding `e.preventDefault()` after focusing is a workaround for FF, so students can play with it by themselves.

RafaelGreen1: > Actually I wouldn't bother changing the code at all (it's a workaround).
> You might want to edit the documentation to mention that adding `e.preventDefault()` after focusing is a workaround for FF, so students can play with it by themselves.

Actually its not a workaround, we don't want the browser to do anything after the focus command
We don't need the browser to follow the link and also to not propagate.
And returning false will work in any browser and it will be forward compatible with any browser changes

avivmu: I looked at [the specs](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid) - A link should make the input focus without any explicit `focus()`.
I think safari handles it correctly - according to [this SO answer](https://stackoverflow.com/a/52902953/2719361) - but I haven't checked it, you are welcome to check it and change the docs accordingly.
We are left with FF - so yeah it _is_ a workaround for a bug (Spec is only what counts).

Regarding `return false` I didn't find a spec so I wouldn't rush to include it in the code.

RafaelGreen1: > I looked at [the specs](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid) - A link should make the input focus without any explicit `focus()`.

Thanks alot for your engagement in this pull request
Could you please quote the specific sentence in the spec that says that
> A link should make the input focus without any explicit `focus()`.
?

bsmth:
Hi both 👋🏻 - it appears the linked bug [is resolved now since Fx 100](https://bugzilla.mozilla.org/show_bug.cgi?id=277178), so I would close this one. If you think this is a mistake, please let me know. Thank you :)

PR #283, opened 2020-08-25
validation-checkbox-disabled.js - remove unnecessary line

RafaelGreen1:
remove unnecessary line of code
instruItem.input.removeAttribute('aria-label');
there is no attribute aria-label to remove


chrisdavidmills:
Yup, good call! I can't remember why I included this in the first place ;-)

PR #284, opened 2020-08-25
Update 7_canvas_walking_animation.html - remove unnecessary calculation

RafaelGreen1:
if we take any whole number and add 2 to it in each iteration it will eventually be a multiple of 13
this sentence from the documentation: “This has to be a multiple of 13 because otherwise the previous code block won’t work, since posX would never equal a multiple of 13!”
is not correct


chrisdavidmills:
Again, thanks a lot for helping me improve the code here. I've updated the associated tutorial also.

RafaelGreen1:
you are welcome, these articles are amazing

chrisdavidmills:
Glad you like them ;-)

PR #285, opened 2020-08-25
Update aria-tabbed-info-box.html - remove unnecessary argument

RafaelGreen1:
the function setTabHandler has only one parameter, but called with two


chrisdavidmills:
Another great change; thanks @RafaelGreen1 !

PR #286, opened 2020-08-28
Update accessibility assessment-finished marking-guide.md considerations

RafaelGreen1:
add consideration - add aria-live to the comments area

### DIFF
--- a/accessibility/assessment-finished/marking-guide.md
+++ b/accessibility/assessment-finished/marking-guide.md
@@ -88,6 +88,7 @@ The overall mark awarded is out of 37. Work out their final mark, and then divid
     <ul>
       <li>Add client-side validation to the comment form</li>
       <li>Use media queries to make the site more usable on mobile devices.</li>
+      <li>Use <code>aria-live</code> to inform screenreader users when the area of the comments is updated.</li>
     </ul>
   </dd>
 </dl>


chrisdavidmills:
@RafaelGreen1 This is a nice idea, but I'm not convinced this is necessary — if the site were dynamic, polled the server, and then automatically inserted new comments that other other users made via XHR/Fetch, then it might be a good idea to inform the user that a new comment has appeared. But even then, it would rely on them having the site open in an active tab, and being on there for quite some time, for it to be useful.

As it is, it is a basic static site, so this wouldn't happen, unless the user inserted a comment of their own, in which case they know they've just inserted a comment — so do they need aria-live to make them aware of this?

RafaelGreen1:
I tried running it with NVDA and after clicking the "show comments" nothing happened (In terms of sound).
I would expect it to tell me about the new form that just has been created.

chrisdavidmills:
Oh, I get you now — the comments were hidden, but you are trying to get them annouced to screenreaders when displayed. And does aria-live achieve this? Also, is it possible to provide an alert along the lines of "2 comments available", rather than it reading out the entire comments block immediately, as that might be annoying.

RafaelGreen1:
correct, after second thought I think that role="alert" is the appropriate solution to this issue (right?)
(I'm pretty new to all these stuff)
about the "2 comments available" they (the comments) are already in a list so screen readers should read out "list with two items" or something like that (no?)

chrisdavidmills:
OK, I've done a bit of testing on this, and I've come to the conclusion that the best low-effort solution here would be to put the attribute on the `<h2>`, and update its text like so:

```
<h2 role="alert">Comments (1 available)</h2>
```

This way when you press the buton it reads out "Comments, 1 available".

If you put it on the comments container, it immediately reads out the entire contents, which is pretty annoying.

Can you update it again, to provide a little more detail like this?

RafaelGreen1:
sure, updated

chrisdavidmills:
Yup, that looks good. I'll merge this now. Thanks for your help here.

PR #288, opened 2020-08-29
Update contact-search.html

Tao93:
if the 'contact not found' code is inside the for loop, for each contact not the one we want, the text 'Contact not found.' would be shown, even if the contact we want is found latterly. 
Though the text 'Contact not found.' is shown for a very short time as long as finally we found the contact we want. But the logic could be improved.

### DIFF
--- a/javascript/building-blocks/loops/contact-search.html
+++ b/javascript/building-blocks/loops/contact-search.html
@@ -26,14 +26,17 @@
       const searchName = input.value.toLowerCase();
       input.value = '';
       input.focus();
+      let found = false;
       for(let i = 0; i < contacts.length; i++) {
         let splitContact = contacts[i].split(':');
         if(splitContact[0].toLowerCase() === searchName) {
           para.textContent = splitContact[0] + '\'s number is ' + splitContact[1] + '.';
+          found = true;
           break;
-        } else {
-          para.textContent = 'Contact not found.';
         }
+      }
+      if (!found) {
+        para.textContent = 'Contact not found.';
       }
     });
       


RafaelGreen1:
actually, it's not true
try the original code with:
```
    let contacts = ['Chris:2232322', 'Sarah:3453456', 'Bill:7654322', 'Mary:9998769', 'Dianne:9384975'];
    for (let i = 0; i < 10000000; i++) {
      contacts.push('Chris:2232322');
    }
    contacts.push('Tao93:2232322');
```
search Tao93 and see what happen

Tao93:
@RafaelGreen1  Thank you for your comment! I tried your long list of contacts, and found you were true. Previously I added a breakpoint at line `para.textContent = 'Contact not found.'`, and found the text in UI would firstly became `Contact not found.` and then became the found result. So, could you tell my why breakpoints are different from a busy loop? Thanks!

Tao93:
Close the PR as @RafaelGreen1 pointed that UI behavior is consistent.

RafaelGreen1:
Hi @Tao93, 
You are welcome.

What I'm going to write below is just my opinion (no resources)
I think it's because how the browser rendering works.
It's not efficient to render the DOM tree for every line of code execution.
But, when you put breakpoints it does render the DOM when you encounter the breakpoint.

keep learning and suggest improvements
Thanks,

Tao93:
Hi @RafaelGreen1 , thank you for your great example and inspiring explanation! I'm a beginner in web development and is going through the MDN learning area.

PR #290, opened 2020-09-06
Update number-guessing-game-start.html - newline

dedeogluhu:
Add new line to make a bit more readable


chrisdavidmills:
That makes sense. Thanks for the contribution!

PR #291, opened 2020-09-14
Update object-basics1-download.html

NatashaMoon:
Changed object member catName to cat.name as it was incorrect format

### DIFF
--- a/javascript/oojs/tasks/object-basics/object-basics1-download.html
+++ b/javascript/oojs/tasks/object-basics/object-basics1-download.html
@@ -46,7 +46,7 @@
     let para1 = document.createElement('p');
     let para2 = document.createElement('p');
 
-    para1.textContent = `The cat's name is ${ catName }.`;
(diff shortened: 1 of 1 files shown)


RafaelGreen1:
Hi @NatashaMoon 
I think it's not a mistake in this case

in the first task, it's written:
> Store the value of the name property inside the **catName** variable, using bracket notation.

so, we use ``` catName ``` to check that the student indeed stored ``` cat.name ``` into the ``` catName ``` variable

chrisdavidmills:
@RafaelGreen1 yup, you are correct. See the first bullet under https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Test_your_skills:_Object_basics#Object_basics_1.

@NatashaMoon I'm going to close this, as it is not needed, but please don't let this put you off — we really appreciate anyone trying to participate and make this resource better.

All the best.

PR #293, opened 2020-09-20
removing redeclared line from events skill test

RafaelGreen1:
following this change: 1d935804999146341b122a79073191152d883159
need to also change the download and marking files


chrisdavidmills:
Lovely, thank you!

PR #294, opened 2020-09-23
Fixed a really small typo

silvercub:
Hi, in the comment on line 31 I changed "commmon" to "common". I'm loving the front end developer path so far. I hope it will lead to a career change down the road so thank you for putting this stuff together.


chrisdavidmills:
Ah lovely, thanks for the fix @silvercub !

And I am so glad you are finding the learning pathway useful ;-)

PR #296, opened 2020-09-24
Update gulpfile.js

RafaelGreen1:
this didn't work for me, so I'm changing it to a working version
also, by gulp's docs, the task API isn't the recommended pattern anymore
https://gulpjs.com/docs/en/api/task/


chrisdavidmills:
Thanks for looking over these @RafaelGreen1 , I really appreciate it. It has been a couple of years since I wrote this material, so I'm not surprised some of it is out of date.

PR #301, opened 2020-10-25
Update animals-table-fixed.html

jjc98375:
Innovated the Animal table in a new column and table to better visualize each term of animals by gender.

### DIFF
--- a/html/tables/basic/animals-table-fixed.html
+++ b/html/tables/basic/animals-table-fixed.html
@@ -1,39 +1,40 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
     <title>Animals table</title>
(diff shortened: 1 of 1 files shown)


RafaelGreen1:
Hi @jjc98375 
Your table looks good and I appreciate your effort to improve the docs but the table is shown under the section:
**Allowing cells to span multiple rows and columns**
So, I think we don't want to remove the `collspan` and `rowspan` attributes.
About the content, we want the male and female names of Hippopotamus (and Crocodile) to be the same to demonstrate the using of `collspan` and `rowspan`
About using `<thead>` and `<tbody>`, it's not explained until the next article [HTML Table advanced features and accessibility](https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Advanced) so it's better to not use in the basic article.

PR #303, opened 2020-10-31
Update index.js

maxganiev:
removing redundant code (lines 155 - 157)


chrisdavidmills:
I'd rather not make this change. I appreciate that the version you are supplying is more compat than the one I've used, but I also think it is harder to parse — I deliberately wanted to keep the different steps separate so that it was easy to see them all.

Thanks anyway.

PR #305, opened 2020-11-02
Update math1-download.html

Theodore354:
the code has been added


chrisdavidmills:
This is supposed to be the starter file for the exercise, so it doesn't need a solution filled in on it.

PR #309, opened 2020-11-11
Changed to the starting file state

BassOfBass:
This file was linked from [HTML accessibility assessments 4](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/Test_your_skills:_HTML_accessibility#HTML_accessibility_4).


chrisdavidmills:
Ah, I see — you are making the downloadable starter file consistent with the live example?

Good fix, thanks for spending the time doing this!

PR #310, opened 2020-11-13
Turned into the starting state

BassOfBass:
As per [example in the task brief](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics/Test_your_skills:_WAI-ARIA#WAI-ARIA_2).


chrisdavidmills (approved):
Good catch, thanks @BassOfBass !

PR #311, opened 2020-11-14
keypress event changed to keydown

ruippeixe:
Since the *keypress* event has been deprecated I think it's better to use *keydown* event.

(I'm not sure if I'm totally right, I just stumbled upon that while learning. It's also my first pull request and couldn't find the rules I should follow, sorry.)


avivmu:
Thanks a lot :+1: 
BTW it's helpful to link to the relevant MDN page.
@chrisdavidmills merge please

Issue #312, opened 2020-11-22
Good/Bad `for` loop examples in "Javascript" article of "Cross-browser testing" section


BassOfBass:
The [examples in this part](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/JavaScript#Fixing_general_JavaScript_problems) state:
> When clicked, each one should alert a message containing its number (the value of i at the time it was created), however each one reports i as 11, because for loops do all their iterating before nested functions are invoked.

However both examples invoke `i` equal to the index of the paragraphs.
Looks like [it was using `var`'s](https://github.com/mdn/learning-area/commit/7c60d316c0c60bcdf1cf0c8d2109e8311e9ac702#diff-4982e743fade2ad25e68695124534f37107dcfe07facd9ea141e20e031edb7c8) in the past, but the example relies on `var` not being block-scoped on top of being assigned to global scope, both of which aren't applicable to `let`.

RafaelGreen1:
just mentioning
in IE it indeed reports i as 11

BassOfBass:
It's fixed now.

hamishwillee:
Thanks @BassOfBass - good spot. As an initial fix I have reverted the `let` into `var`: https://github.com/mdn/learning-area/pull/313. 

Reopening as it isn't quite fixed; generally speaking people should and will will use `let` rather than `var` so even though the text is no longer wrong, it isn't complete. I'm trying to tidy now and I will get back to you for review shortly.

hamishwillee:
The desirable thing to do would be to recommend that people simply use `let` rather than `var` for the iteration variable. However because of IE11 not working properly (as pointed out by @RafaelGreen1 ) I don't think we can do that.

So what I have done is kind of gone "half way". The doc now makes it clear that this is mostly a problem with iterating using a global variable. At the end though it points out that `let` should work, and the specific case it does not. My theory being that as IE11 fades into obscurity this will remind us we can eventually update the "good" example.

![image](https://user-images.githubusercontent.com/5368500/99930918-e1bccd00-2da6-11eb-888c-9b77bfaae486.png)

The alternative was to update this to recommend `let` now, but I think it is too early. 

Does this seem reasonable @RafaelGreen1 @BassOfBass ?

BassOfBass:
Pretty sure the good example should also run in a "global" IIFE to avoid any global scoping issues.
Something like this:
```javascript
;(function(){
  ...
  ...
  ...
})();
```
[...]

PR #313, opened 2020-11-23
bad-for-loop: revert to var

hamishwillee:
This got turned from `var` to `let` at some point, which unfortunately (for the docs) turns it from a bad example into a good one. As discussed in https://github.com/mdn/learning-area/issues/312, using `let` scopes the value passed to the function so the example works. With `var` the function is called using the final global value of the variable, which is 11.


hamishwillee:
Self merging - note, more work to "properly" fix up the docs, but this means they are no longer incorrect.

PR #323, opened 2020-12-19
Viewport

fikstoday:
Viewport meta tag added to all html files under APIS folder. Those with meta tag updated by including initial-scale=1.0


chrisdavidmills (approved):
Brilliant, thank you @fikstoday !

PR #324, opened 2021-01-03
Fix typo: advanced-text2-download.html

TomoiagaPaul:
Fix typo: letter 'O' instead of number '0'


chrisdavidmills:
What is there is correct — The "O" stands for a single oxygen atom.

PR #329, opened 2021-01-12
Added max-width

fikstoday:
1. Resized h1 icons
2. Added a grid to body div
3. Aligned header, aside, main and footer as grid rows stacked

### DIFF
--- a/javascript/apis/fetching-data/can-store/can-style.css
+++ b/javascript/apis/fetching-data/can-store/can-style.css
@@ -175,3 +175,30 @@ section p {
   background: black url(icons/soup.png) no-repeat 2px 2px;
   background-size: 24px 24px;
 }
+
+/* mobile responsive styles*/
+@media screen and (max-width: 550px){
+  body, body div {
+    display: grid;
+    max-width: 400px;
+  }
+  header {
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
While adding some simple responsive styles to this example isn't unreasonable, I don't have time to test this, and the point of this example is not to demonstrate responsive techniques, so I'd argue that this isn't really needed.

fikstoday:
Acceptable in this regard, please fell free to share more appropriate repos.
________________________________

chrisdavidmills:
Actually @fikstoday , I feel really mean for not reviewing this. I'm going to make some time now.

chrisdavidmills (changes requested):
This is looking mostly good, nice work @fikstoday !

As you can see from the screenshot, there are a couple of small fixes to make before we can merge this:

1. There is an issue with the border on the right hand side of the box containing the filter controls.
2. The search input could do with being made the same width as the other form elements.
3. In some cases, the product name overlaps the product icon. Maybe on the narrow screen layout, you could just hide the icons? Up to you anyway.

Screen Shot 2021-01-12 at 9 05 42 PM

PR #330, opened 2021-01-12
Correct aside border

fikstoday:
1. Removed right border for aside element
2. Added bottom border for aside
3. Introduced margin top for header, margin bottom for footer
4. Resized input width, reset left & right padding
5. Removed food item __icons__


chrisdavidmills (approved):
Looking good! I made a couple of other small changes (e.g. giving the input element some left padding), and I think this is ready to merge now. Thanks @fikstoday !

PR #332, opened 2021-02-01
Update contact-search.html

tankris:
Changed the else statement to be an else if statement that instead of being 
executed for every array element that isn't a match, is executed only upon 
not finding a matching entry in the entire contacts array

### DIFF
--- a/javascript/building-blocks/loops/contact-search.html
+++ b/javascript/building-blocks/loops/contact-search.html
@@ -31,7 +31,7 @@
         if(splitContact[0].toLowerCase() === searchName) {
           para.textContent = splitContact[0] + '\'s number is ' + splitContact[1] + '.';
           break;
-        } else {
+        } else if(i === contacts.length - 1) {
           para.textContent = 'Contact not found.';
         }
       }


chrisdavidmills:
This change does make sense; thanks for the suggestion @tankris 

Can you also update the tutorial page where this example is used, so the code and description matches the code in this repo?

The page section is at https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Looping_code#exiting_loops_with_break

 And the source code for the tutorial is at https://github.com/mdn/content/blob/main/files/en-us/learn/javascript/building_blocks/looping_code/index.html

tankris:
Hi! I shall make the necessary change there as well.

Also I corrected the grammatical error in the commit message. Hence I forced pushed the update. Hope that isn't an issue

chrisdavidmills:
> Hi! I shall make the necessary change there as well.
> 
> Also I corrected the grammatical error in the commit message. Hence I forced pushed the update. Hope that isn't an issue

Not a problem at all; thanks @tankris !

chrisdavidmills (approved):
OK, let's get this in. Thanks!

PR #339, opened 2021-02-16
Update planets-data.html

atakndogn:
Scope tag for "Dwarf Planets" should be "rowgroup" instead of "row" as "Dwarf planets" is a group name for all "dwarf" ones, not only for"Pluto".


chrisdavidmills (approved):
Yup, I'd agree with that; thanks for your help @atakndogn !

PR #348, opened 2021-03-17
Update `myAge` variable

Yuniac:
Hi, The number in the variable `myAge` is being stored as a string and not as a number. Thus, when we use it in the template literal later on the result is concatenating the strings together and not mathematically adding the numbers together.

Current state: '42' + 20 and its outputting like this: `4220`

The proposed change: 42 (a number now, no longer a string) + 20 will output as it's intended to and results in `62`, Chris's age :)

### DIFF
--- a/javascript/introduction-to-js-1/tasks/variables/variables3-download.html
+++ b/javascript/introduction-to-js-1/tasks/variables/variables3-download.html
@@ -31,7 +31,7 @@
     const myName = 'Default';
     myName = 'Chris';
 
-    let myAge = '42';
+    let myAge = 42;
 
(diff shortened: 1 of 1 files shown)


RafaelGreen1:
Hi @Yuniac 
Thanks for your pull request but I think the change is not needed in this case.
In the task it's written:

> The final task for now — in this case you are provided with some existing code, which has **two errors** present in it

So, I think the error is there for a purpose and no change is needed here.
Thanks,

Yuniac:
Hello, thanks for replying to me.
You are right, I forgot that the sample code there contains two errors on purpose, sorry.

PR #350, opened 2021-03-19
Give each checkbox the same 'name'

jrsnzc:
When handling multiple related items.

### DIFF
--- a/html/forms/native-form-widgets/checkable-items.html
+++ b/html/forms/native-form-widgets/checkable-items.html
@@ -13,23 +13,23 @@
           <ul>
             <li>
               <label for="carrots">Carrots</label>
-              <input type="checkbox" checked id="carrots" name="carrots" value="carrots">
+              <input type="checkbox" checked id="carrots" name="vegetable" value="carrots">
             </li>
             <li>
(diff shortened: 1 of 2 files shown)


chrisdavidmills (approved):
I agree with you, this seems to be the right way to handle related checkbox items.

I'd also like to ask — would you be willing to help update the related pages on MDN that talk about checkboxes?

We ought to mention this on https://developer.mozilla.org/en-US/docs/Learn/Forms/Basic_native_form_controls (which I think contains the vegetables example)

We already mention this best practice at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes, which is good.

jrsnzc:
Hello @chrisdavidmills,
Yes, it would be my pleasure :-)

chrisdavidmills:
> Hello @chrisdavidmills,
> Yes, it would be my pleasure :-)

That is great, thank you! I would recommend that you search the site to see where else checkboxes are mentioned, and think about where it might be appropriate to mention this best practice. Don't hesitate to ask if you need any help.

PR #352, opened 2021-03-22
Add fix and talk about multiple related checkboxes

jrsnzc:


chrisdavidmills (approved):
This also looks great; nice work @jrsnzc !

PR #361, opened 2021-05-04
Reduce slice function

kemonprogrammer:
Since the string.slice() function allows negative numbers as arguments, it can be written shorter.


kemonprogrammer:
But I can understand if the longer version was used, to show that you can concatenate string methods.

PR #370, opened 2021-06-17
Fixed errors in the task: Test your skills: Events

nitramrelpmur:
Fixed some errors in the task: [Test your skills: Events](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Test_your_skills:_Events)

- The script element was outside the body element. This resulted in a validation error.
- There is also a warning because the section element lacks a heading. I therefore inserted a placeholder that is not displayed.

### DIFF
--- a/javascript/building-blocks/tasks/events/events1-download.html
+++ b/javascript/building-blocks/tasks/events/events1-download.html
@@ -23,12 +23,11 @@
 
   <body>
 
-    <section class="preview">
+    <section class="preview"><h2 style="display:none">Placeholder</h2>
     </section>
 
     <button class="off">Machine is off</button>
 
-  </body>
   <script>
     let btn = document.querySelector('.off');
 
@@ -37,4 +36,5 @@
 
   </script>
 
+  </body>
 </html>
--- a/javascript/building-blocks/tasks/events/events1-download.html
+++ b/javascript/building-blocks/tasks/events/events1-download.html
@@ -23,7 +23,7 @@
 
   <body>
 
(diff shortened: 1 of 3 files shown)


javascript/building-blocks/tasks/events/events1-download.html
```diff
@@ -23,12 +23,11 @@
 
   <body>
 
-    <section class="preview">
+    <section class="preview"><h2 style="display:none">Placeholder</h2>
```
Rumyra, line 26:
```suggestion
    <section class="preview">
```

javascript/building-blocks/tasks/events/events2-download.html
```diff
@@ -22,15 +22,13 @@
 
   <body>
 
-    <section class="preview">
+    <section class="preview"><h2 style="display:none">Placeholder</h2>
```
Rumyra, line 25:
```suggestion
    <section class="preview">
```

javascript/building-blocks/tasks/events/events3-download.html
```diff
@@ -27,7 +27,7 @@
 
   <body>
 
-    <section class="preview">
+    <section class="preview"><h2 style="display:none">Placeholder</h2>
```
Rumyra, line 30:
```suggestion
    <section class="preview">
```

Rumyra:
Hi @nitramrelpmur thanks so much for your contribution! I agree the script tags should be within the body, however I think the addition of headers (even tho valid) with a display non attribute would be confusing for peeps undertaking this exercise 👍

PR #375, opened 2021-06-22
Update math1-download.html

iamwolf32:


wbamberg:
Thanks for the PR, but completed exercises should not be merged here - this file is the "empty" file for people to download and try the exercise.

PR #379, opened 2021-06-28
Update flexbox0.html

ammelnaggar:


Rumyra:
Hi @ammelnaggar thanks for the contribution, however this example needs to stay the same to be consistent with the lesson on the mdn page :)

PR #381, opened 2021-07-12
Information text corrected to match html.onclick()

johnjinto:
Please consider the following small improvement made in information text to make it more clear since clicking inside the empty list also will add the 1st item.

Information text _"Click anywhere outside list"_ changed to _"Click anywhere on the page"._ 

Hope this description will more accurately indicate the functionality of **html.onclick()**

PR #382, opened 2021-07-12
Revert "Information text corrected to match html.onclick()"

johnjinto:
Reverts mdn/learning-area#381

Thank you!

But now we have to also modify the image to align with JS code. 

image

How can we do this? Requesting @chrisdavidmills to help change the image to maintain consistency.


teoli2003:
To fix this, you should make a new screenshot and open PR in mdn/content to replace this file by it:

https://github.com/mdn/content/blob/main/files/en-us/learn/html/introduction_to_html/the_head_metadata_in_html/js-and-css.png

PR #383, opened 2021-07-14
add comment and more example for new learner 

danielpm307:


danielpm307:
new href attrb added

Rumyra:
Hi @danielpm307 thanks for your contribution. We're currently working on updating the learning area, so this will change in due course. Sorry for taking time to get back to you :)

PR #386, opened 2021-07-26
Update marking.md

MadhavKwatra:
fixing a typo in task 2. Replacing textarea with select.The task was about select and not textarea so changing id of select.


hamishwillee (approved):
Nice catch. Thank very much

PR #389, opened 2021-08-02
Update flexbox-wrap0.html

Sabbu754:


Rumyra:
Hi @Sabbu754 thanks so much for the contribution, but this example needs to stay the way it is so learners can add the correct properties themselves 👍

PR #397, opened 2021-09-03
credit card expiry date format

Biki-das:


hamishwillee:
@Rumyra Where does the file being edited here come from? Is it an itereactive example or something? I was assuming we were working on https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form

@Biki-das The approach used in the example is inconsistent. The rest of the example/document uses HTML attribute for formatting. This PR then adds javascript to test the input, and on just this element (not everything). So while I appreciate that using JavaScript can create a much better experience it doesn't really work for a topic about learning how to use the HTML we forms. Does that make sense.

What I was envisaging was a simple change like below. Note that now we have `required` as part of the attributes, and also the pattern.  

```
            <p>
              <label for="expiration">
                <span>Expiration date:</span>
                <strong><abbr title="required">*</abbr></strong>
              </label>
              <input type="text" name="expiration" required="true" placeholder="MM/YY" pattern="^(0[1-9]|1[0-2])\/([0-9]{2})$">
            </p>
```

Biki-das:
> @Rumyra Where does the file being edited here come from? Is it an itereactive example or something? I was assuming we were working on https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form
> 
> @Biki-das The approach used in the example is inconsistent. The rest of the example/document uses HTML attribute for formatting. This PR then adds javascript to test the input, and on just this element (not everything). So while I appreciate that using JavaScript can create a much better experience it doesn't really work for a topic about learning how to use the HTML we forms. Does that make sense.
> 
> What I was envisaging was a simple change like below. Note that now we have `required` as part of the attributes, and also the pattern.
> 
> ```
>             <p>
>               <label for="expiration">
>                 <span>Expiration date:</span>
>                 <strong><abbr title="required">*</abbr></strong>
>               </label>
>               <input type="text" name="expiration" required="true" placeholder="MM/YY" pattern="^(0[1-9]|1[0-2])\/([0-9]{2})$">
>             </p>
> ```

well the experience should be realistic in my opinion and i can create a .js file to link the js, if we need to use certain stuff for something , we have to use it rather caring about incosistency as we need it here

hamishwillee:
> well the experience should be realistic in my opinion and i can create a .js file to link the js, if we need to use certain stuff for something , we have to use it rather caring about inconsistency as we need it here

The rest of the document is about HTML attributes and now, suddenly you're introducing JavaScript. That's basically saying "you don't need to learn about HTML attributes". But the point here is that you do! What if someone turns off the JavaScript in the browser - a lot of people do? Now your code doesn't work.

In any case though, you're teaching the information in chunks, building towards expertise. In this page its about attributes. If you jump ahead, then your students miss out of what the document is trying to teach them.

Biki-das:
> > well the experience should be realistic in my opinion and i can create a .js file to link the js, if we need to use certain stuff for something , we have to use it rather caring about inconsistency as we need it here
> 
> The rest of the document is about HTML attributes and now, suddenly you're introducing JavaScript. That's basically saying "you don't need to learn about HTML attributes". But the point here is that you do! What if someone turns off the JavaScript in the browser - a lot of people do? Now your code doesn't work.
> 
> In any case though, you're teaching the information in chunks, building towards expertise. In this page its about attributes. If you jump ahead, then your students miss out of what the document is trying to teach them.

oh got it, thanks for teaching me , now how should i approach

Rumyra:
I wholeheartedly agree with @hamishwillee here. This is a learning document focussed on HTML attributes for forms and while the JavaScript may be 'real world' it would be confusing for learners to experience it in this way. As a once teacher the rule was 'You can't tell the whole truth at the beginning' however much you would like to :)

hamishwillee:
>oh got it, thanks for teaching me , now how should i approach

Hmm. I think the first thing would be for you to see if you can test my suggestion and confirm it works on a couple of browsers. 
I think it should be OK, but having you confirm it is sensible would be great.

Then update https://github.com/mdn/learning-area/blob/master/html/forms/html-form-structure/payment-form.html as you have here.
 
Then copy it into https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form in the same place you did your first PR - i.e. 
image

Does that make sense?

loversun08:
yzkt

Biki-das:
> > oh got it, thanks for teaching me , now how should i approach
> 
> Hmm. I think the first thing would be for you to see if you can test my suggestion and confirm it works on a couple of browsers.
> I think it should be OK, but having you confirm it is sensible would be great.
> 
> Then update https://github.com/mdn/learning-area/blob/master/html/forms/html-form-structure/payment-form.html as you have here.
> 
> Then copy it into https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form in the same place you did your first PR - i.e.
> image
> 
> Does that make sense?

thanks @hamishwillee , for teaching me, i learned something new this time and yes i am testing your sugesstion and will push PR regarding this

Biki-das:
Pulled another PR  #400 , @hamishwillee , hope you review it and again thanks for guiding me out

hamishwillee:
Hi @Biki-das 
Thanks for that - looks great - I have merged. I'm going to close _this_ PR.

next step is to update https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form in line with this change.

Biki-das:
Thanks for that quick review, will be updating that one too.

PR #398, opened 2021-09-04
fix typo in letterheaded paper marking guide

mikoMK:


wbamberg (approved):
👍 thank you for the fix, @mikoMK !

PR #406, opened 2021-09-18
Replace let with const in "Silly story generator"

mikoMK:
Only the "newStory" variable needs let. Others can be changed to const.

I tested the new code and it still runs correctly.


wbamberg (commented):
Sorry to be so slow looking at this PR @mikoMK .

This is mostly taken care of in https://github.com/mdn/learning-area/pull/427/files#diff-20970d5f178e825531b2641722e8c3d4ba6704342451db85f00642f30bb29488 , except I missed the updates to the marking guide and raw-text.

If you would like to push those updates to https://github.com/mdn/learning-area/pull/427, that would be great, otherwise I will (I didn't want to just do it without asking because this is your work).

mikoMK:
No need to be sorry. I know you were focused on the Markdown conversion project, so I didn't want to nag you about my small PRs. 😉 

Could you please incorporate those change to #427? I don't know how I could do that.

wbamberg:
> Could you please incorporate those change to #427? I don't know how I could do that.

I just added those changes. Thanks again for this PR!

wbamberg:
> I don't know how I could do that.

I was thinking you could push to my PR, but I guess you can't unless your a maintainer for this repo.

PR #408, opened 2021-09-19
Improve loops marking guide

mikoMK:
We ask the student to use three different types of loops. We should do the same in the marking guide. Therefore I changed do...while to while in task 2. After this change we have:
* Task 1: for
* Task 2: while
* Task 3: do...while

I tested the changes in an online editor.


wbamberg (commented):
Sorry to be so slow to look at this PR.

I think this is going to be taken care of by https://github.com/mdn/learning-area/pull/421, which adds a `for...of` loop and also ensuresa they all use a different loop type.

PR #409, opened 2021-09-30
Update basic-text1-download.html

xxaviermainville:


Rumyra:
Hi @xxaviermainville thanks so much for the pr, however this example specifically should not have any text formatting, so learners can add it themselves - thanks :)

PR #414, opened 2021-10-22
Update basic-for.html

Bellusfennec:


javascript/building-blocks/loops/basic-for.html
```diff
@@ -14,7 +14,7 @@
 
     <script>
     const cats = ['Bill', 'Jeff', 'Pete', 'Biggles', 'Jasmin'];
-    let info = 'My cats are called ';
+    let info = 'My cats are called: ';
     const para = document.querySelector('p');
```
maksliudva, line 18: deep work

hamishwillee:
Thank you. I'm closing this because it doesn't help a user better understand the material and I don't think it is worth the extra time required to update the corresponding docs: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Looping_code

PR #417, opened 2021-11-07
8_canvas_drawing_app: Add a visible default value

hamishwillee:
Fixes https://github.com/mdn/content/issues/10338

The default value for an input selector is black, which means that the drawing app starts as black on black - a little confusing for users. This sets the initial colour as red.


teoli2003 (approved):
Sounds good!

teoli2003:
cc/ @Rumyra as I don't have writing rights in this repo.

PR #418, opened 2021-11-12
Add a new task to the JS Arrays section

wbamberg:
Part of https://github.com/mdn/content/issues/10337 .

I'm working on adding some bits to the ["Arrays" page of the JS Learning area](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Arrays).

In particular I wanted to add bits on `slice()`, `indexOf()` and array iteration methods including `for...of`, `map()` and `filter()`.

So I thought it would be worth adding a "test your skills" section to the [existing tasks](http://localhost:5000/en-US/docs/Learn/JavaScript/First_steps/Test_your_skills:_Arrays). The exercise will be described something like:

> For this array task, we provide you with a starting array listing the names of some birds.
>
> * Find the index of the "Eagles" item, and use that to remove the "Eagles" entry.
> * Make a new array from this one, called `eBirds`, that contains only birds from the original array whose names begin with the letter "E". Note that [`startsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) is a great way to check whether a string starts with a given character.
>
> If it works, you should see "Egrets,Emus" appear in the page.

### DIFF
--- a/javascript/introduction-to-js-1/tasks/arrays/marking.md
+++ b/javascript/introduction-to-js-1/tasks/arrays/marking.md
@@ -14,7 +14,7 @@ Last of all, add an item to the start of the array using `.unshift()`.
 
 The answer should look something like this:
 
-```
+```js
 let myArray = ['cats', 'dogs', 'chickens'];
 
 myArray[0] = 'horses';
@@ -22,7 +22,6 @@ myArray[1] = 'pigs';
 
 myArray.unshift('crocodiles');
 ```
-
 
 ## Task 2
 
@@ -36,7 +35,7 @@ Last up, retrieve the last item in the array with `myArray[myArray.length - 1]`;
 
 The example should look something like this:
 
-```
+```js
 let myString = 'Ryu+Ken+Chun-Li+Cammy+Guile+Sakura+Sagat+Juri';
 
 let myArray = myString.split('+');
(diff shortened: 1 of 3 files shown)


wbamberg:
@ddbeck , you asked me to ping you in a couple of these :).

javascript/introduction-to-js-1/tasks/arrays/marking.md
```diff
@@ -70,4 +69,24 @@ myArray.forEach(function(element, index) {
...
+The function passed to `filter()` can use `bird.startsWith()` to test whether the bird starts with "E".
+
+```js
+const eaglesIndex = birds.indexOf("Eagles");
+birds.splice(eaglesIndex, 1);
+
+function startsWithE(bird) {
+  return bird.startsWith("E");
+}
+const eBirds = birds.filter(startsWithE).sort();
```
ddbeck, line 91: Just to prove I'm paying attention: is the sort obligatory? Since the array is already sorted, I'm not sure if it adds much to include it.

ddbeck (approved):
Looks good! One optional idea in a line comment, but you're free to merge without it.

javascript/introduction-to-js-1/tasks/arrays/marking.md
wbamberg: No, you're right, I don't know what I was thinking there. Looking at this PR I saw some other problems. There are two versions of each task, one that's supposed to run in the page and one that's supposed to be downloaded. These should stay in sync, and they weren't.

Also more `let`->`const`.

ddbeck (approved):
That's a bit tidier. Thank you! 🎉

PR #421, opened 2021-11-20
Update JS Loop tasks to include for...of

wbamberg:
Part of https://github.com/mdn/content/issues/10337.

This PR updates the tasks for the [Looping code](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Looping_code) part of the JS learning area to include `for...of`. I've also changed another one to `for` just so that form is represented too (the idea is that each task should use a different loop).

I also updated one of the example pages to match changes made in https://github.com/mdn/content/pull/10645.

### DIFF
--- a/javascript/building-blocks/loops/contact-search.html
+++ b/javascript/building-blocks/loops/contact-search.html
@@ -26,15 +26,17 @@
       const searchName = input.value.toLowerCase();
       input.value = '';
       input.focus();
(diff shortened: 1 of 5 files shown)


lucalves (approved):

wbamberg:
@ddbeck , you asked me to ping you in a couple of these :).

PR #423, opened 2021-11-21
The value 255 was excluded

yusif-a:
Return value of random doesn't include number itself, so random(255) never generates 255.


hamishwillee (approved):
Thanks. Good catch. I fixed up the docs that reference this in https://github.com/mdn/content/pull/10686

PR #425, opened 2021-12-01
Add a new Functions task

wbamberg:
This is part of https://github.com/mdn/content/issues/10337.

https://github.com/mdn/content/pull/10799 updates the [JS Learning Area's "Functions" page](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions), mostly to talk about arrow functions.

There's a ["Test Your Skills" page for functions](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Test_your_skills:_Functions) and I wanted to add a new task to this page to cover arrow functions. This PR updates the code examples for the new task. The new task is just to convert a `filter()` parameter from a named function statement to an arrow function expression.

To go along with this PR will be a PR to update https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Test_your_skills:_Functions to describe the new task, which will say something like:

> In this task, we have an array of names, and we're using `Array.filter()` to get an array of only names shorter than 5 characters. The filter is currently being passed a named function `isShort()` which checks the length of the name, returning `true` if the name is less than 5 characters long and `false` otherwise. We'd like you to change this into an arrow function. See how compact you can make it.

### DIFF
--- a/javascript/building-blocks/functions/first.js
+++ b/javascript/building-blocks/functions/first.js
@@ -1,4 +1,4 @@
-let name = 'Chris';
+const name = 'Chris';
 function greeting() {
-  alert('Hello ' + name + ': welcome to our company.');
+  alert(`Hello ${name}: welcome to our company.`);
 }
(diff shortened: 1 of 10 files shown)


Elchi3 (approved):
Nice! 👍

PR #433, opened 2021-12-20
"em" no longer in code on MDN

Tman1314:
The  "em" portion of the code for the "Payment Form" is no longer used, making the CSS code for it unneeded. The code seems to use placeholder instead of an "em" tag in the label.


hamishwillee:
Thanks @Tman1314 - good cleanup.

PR #440, opened 2022-01-10
Update index.html

glmvc:
One `<br>` tag was written like a closing tag


wbamberg (approved):
👍 nice catch, thank you @glmvc !

PR #446, opened 2022-01-28
Update main.js

IblackdI:
In the article, it implies that function randomRGB()  is already in main.js and you don't need to add it.


wbamberg (approved):
Thank you @IblackdI ! You are totally right.

PR #447, opened 2022-01-30
added solution

lokicodedaily:
add a solution for this loop file


hamishwillee:
Hi @lokicodedaily
Thanks, but these are templates that people are supposed to try out with their own solutions  - so we definitely do not want to put in the "right answer" here.

PR #448, opened 2022-01-30
loop2 solution added

lokicodedaily:


hamishwillee:
Closing. Thanks but this is a template that works as-is with the documentation. We don't want the solution here.

PR #449, opened 2022-02-02
Update strings1-download.html

willsan0723:
This breaks as it is because of the ' character in the quote. Used backticks so that this doesn't have errors.


wbamberg:
Thank you for the PR! But I think this breakage is intentional, isn't it? It goes along with http://localhost:5000/en-US/docs/Learn/JavaScript/First_steps/Test_your_skills:_Strings#strings_1 :

> You'll find that you get an error at this point. Can you fix the problem with quoteStart, so that the full quote displays correctly?

PR #450, opened 2022-02-03
Fixed legend attribute.

heppleton:
Legend for "Check shipping details" was set to "contenteditable:true", which is incorrect, so this attribute has been removed.


bsmth (approved):
Makes sense to me, thank you 👍🏻

PR #463, opened 2022-02-25
JS Learn: update client-side APIs, part 5

wbamberg:
This is the learning-area PR corresponding to https://github.com/mdn/content/pull/13268.

Marking as draft because this and https://github.com/mdn/content/pull/13268 should be merged together, and changes in one will need corresponding changes in the other one.

But this is ready for review.

The first commit: https://github.com/mdn/learning-area/pull/460/commits/02d9d16178b0948e3791ac727e2744c1327ea17f keeps all the code the same, but reorganizes it so the JS code is in its own files rather than inside the HTML. I think this is easier to work with.

The next two commits - https://github.com/mdn/learning-area/commit/a8512af780f5c897c9d19aadc046a4fd3301861a and https://github.com/mdn/learning-area/commit/5fcf9a3a48a57da472fd119a87958192072e1c99 - are the actual updates.


Elchi3 (approved):
Nice! All examples are still functional for me.

PR #464, opened 2022-02-25
Update strings1-download.html

Samharakan:


kleners1:
public sealed class String : ICloneable, IComparable, IComparable<string>, IConvertible, IEquatable<string>, System.Collections.Generic.IEnumerable<char>

bsmth:
Thanks a lot - I'm closing this one because it will [fix the problem that other readers should try to solve](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Test_your_skills:_Strings#strings_1), but you have a solution, well done. Closing for now, thank you :)

PR #470, opened 2022-03-01
Replace deprecated event.keyCode with event.key

AnilSeervi:
Corresponds to mdn/content#13328

Review : @wbamberg

### DIFF
--- a/tools-testing/cross-browser-testing/accessibility/fake-div-buttons.html
+++ b/tools-testing/cross-browser-testing/accessibility/fake-div-buttons.html
@@ -47,7 +47,7 @@
       }
 
       document.onkeydown = function(e) {
-        if(e.keyCode === 13) { // The Enter/Return key
+        if(e.key === 'Enter') { // The Enter/Return key
           document.activeElement.onclick(e);
         }
       };


AnilSeervi:
Should we leave the current example to use event handler(`onclick`) rather than using `addEventListener` Listener?

wbamberg (approved):
👍 thank you @AnilSeervi !

wbamberg:
> Should we leave the current example to use event handler(`onclick`, `onkeydown`) rather than using `addEventListener` Listener?

Yes, ideally we would update this example (and all the examples here) to use good practice modern JS, and that would also include things like `for...of` instead of `for`, and consistently using `const`/`let` (although this example seems OK in that respect). As you might know I've been doing this for the JS part of the learning area (https://github.com/mdn/content/issues/10337), but it would be great to have it for the other bits as well.

It is a bit involved because we have to update the code in here and the corresponding code blocks in the mdn/content pages as well.

If you did want to file some PRs to update some of this content I'd be very happy to review them.

Thanks as always for your work on MDN @AnilSeervi !

PR #473, opened 2022-03-03
JS Learn: update client-side APIs, part 7

wbamberg:
This is the learning-area PR corresponding to https://github.com/mdn/content/pull/13467.

I think it's not possible to test the service worker example locally, but I set my fork's GitHub page source to this branch, so you should be able to try it out at https://wbamberg.github.io/learning-area/javascript/apis/client-side-storage/cache-sw/video-store-offline/.

One change here worth mentioning is that I removed the `window.onload()` stuff, because AFAICT using the `defer` attribute in the `<script>` tag should ensure that the script doesn't run until the page is loaded, right?

Marking as draft because this and https://github.com/mdn/content/pull/13467 should be merged together, and changes in one will need corresponding changes in the other one.

But this is ready for review.

### DIFF
--- a/javascript/apis/client-side-storage/cache-sw/video-store-offline/index.js
+++ b/javascript/apis/client-side-storage/cache-sw/video-store-offline/index.js
@@ -1,24 +1,23 @@
-window.onload = function() {
-  // Create constants
-  const section = document.querySelector('section');
-  const videos = [
-    { 'name' : 'crystal' },
-    { 'name' : 'elf' },
-    { 'name' : 'frog' },
-    { 'name' : 'monster' },
-    { 'name' : 'pig' },
-    { 'name' : 'rabbit' }
-  ];
-  // Create an instance of a db object for us to store our database in
-  let db;
+// Create constants
+const section = document.querySelector('section');
+const videos = [
+  { 'name' : 'crystal' },
+  { 'name' : 'elf' },
+  { 'name' : 'frog' },
+  { 'name' : 'monster' },
+  { 'name' : 'pig' },
+  { 'name' : 'rabbit' }
+];
+// Create an instance of a db object for us to store our database in
+let db;
 
 function init() {
   // Loop through the video names one by one
-  for(let i = 0; i < videos.length; i++) {
+  for(const video of videos) {
     // Open transaction, get object store, and get() each video by name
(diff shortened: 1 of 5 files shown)


charlesumesi:
Hi,

I can't get this latest version of your IndexedDB example to work. Occasionally, I get a very delayed update of the list if I re-upload the page, but this is rare. The previous version of the JavaScript code I saw (and saved) back in January (this year) does work.

I have tried identifying the cause of the problem without success. I even tried changing the database and the version number, as well as encapsulating the whole code (excluding the db declaration) within a function (as was the case in the January version) but none of these have worked in the new version. 

There are significant differences between the two versions and so I've been struggling to make meaningful comparisons between both versions to identify the root of the problem. The January version is attached (as a txt file, since js files are not supported by this form).
[index-start.txt](https://github.com/mdn/learning-area/files/9379220/index-start.txt)


Kind regards
PS: And in case you're wondering, yes, I did put the correct name of the js file in the (correct) html file.

teoli2003:
Thank you, @charlesumesi. Can you try something: could you add the `window.onload = function()` { that have been removed? And tell us if this change something?

charlesumesi:
@teoli2003, thanks for getting back to me.

I've just tried what you suggested and it didn't work. Incidentally, I also tried it before I sent my previous post, and it didn't work then, so I'm unsure where the issue is.

teoli2003:
I checked out the git repo, and tried the example in both Firefox and Chrome. It worked when using a local server, and didn't, as expected, when using the `file:` protocol.

Did you use a local server to test this?

charlesumesi:
Hi, unfortunately, I'm not getting anything on any of the browsers even when using a local server (localhost:8000/index_v2.html). And of course, nothing when using the file directly.

teoli2003:
There is no `index_v2.html`. In such cases, you really should try the files as obtained via git, not copied or modified.

charlesumesi:
Hi, I have now tried the version from the link just before the section on _Storing complex data via IndexedDB_ and that does work. All this while, I had been using the version posted in parts across the article preceding the aforementioned section. There might have been something missed (in the article), but I assumed when the parts were added together, you got the same thing as the version at the end of the article. I guess that assertion was wrong!

Thanks for your help and good advice.

PR #479, opened 2022-03-18
Update text-start.html

eichlerFO:


bsmth:
Well done, looks good 🎉 We're not merging in changes from this as it would fix the starting template that other readers should use. Thank you :)

PR #480, opened 2022-03-19
Add exercises to the learing area page

ra2dd:
Add missing exercices from [MDN](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math) so users can solve problems without going back to the MDN page each time.


bsmth:
Hi @ra2dd thanks for the suggestion, there's some unaddressed review comments above, do you plan to come back to this one?

estelle (approved):

Issue #482, opened 2022-03-22
Closing link tag on html file


galopyz:
There is a closing tag for link tag in learning-area/javascript/asynchronous/sequencing-animations/start/index.html.

There is another one in finished directory of sequencing-animations.

PR #483, opened 2022-03-22
Remove link closing tag for 'Closing link tag on html file #482'

galopyz:
Fixing  'Closing link tag on html file #482'


teoli2003 (approved):
Thank you, yes, `<link>` is an empty element and doesn't need a closing tag.

PR #486, opened 2022-04-02
Change replace() to replaceAll() in "Silly story generator" solution

mikoMK:
Fixes mdn/content#14549

@wbamberg As discussed in the above issue


wbamberg (approved):
👍 looks great, thank you @mikoMK ! We should also update https://github.com/mdn/content/blob/ff56fa1d35fc4b414f2ee75ffb67aa22358a116b/files/en-us/learn/javascript/first_steps/silly_story_generator/index.md?plain=1#L118 I guess - let me know if you will, or I can take care of it if you like.

mikoMK:
You're right, thanks. I didn't think about that.
I will take care of it. 👍

PR #489, opened 2022-04-15
Update 'Silly Story Generator' marking guide with replaceAll()

mikoMK:
In PR #486 we changed the solution code from `replace()` to `replaceAll()`.
This PR updates the marking guide accordingly. I removed the part about `replace()` with regex since we never taught it in the lesson and giving two different solutions (`replace()` and `replaceAll()`) seems to be enough.


wbamberg (approved):
👍 thank you @mikoMK !

PR #490, opened 2022-04-15
Remove typemustmatch attribute

mikoMK:
The `typemustmatch` attribute was removed from the spec and subsequently from MDN.
We still use it in two live examples and the marking guide. This PR removes them.

Background: https://github.com/mdn/content/pull/3655

### DIFF
--- a/html/multimedia-and-embedding/other-embedding-technologies/object-image.html
+++ b/html/multimedia-and-embedding/other-embedding-technologies/object-image.html
@@ -8,7 +8,7 @@
     <h1>Object element example</h1>
 
  
-    <object data="dinosaur.jpg" type="image/jpeg" width="400" height="341" typemustmatch>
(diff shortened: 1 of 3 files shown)


wbamberg (approved):
👍 thank you for the fix, @mikoMK !

html/multimedia-and-embedding/other-embedding-technologies/object-image.html
```diff
@@ -8,7 +8,7 @@
     <h1>Object element example</h1>
 
  
-    <object data="dinosaur.jpg" type="image/jpeg" width="400" height="341" typemustmatch>
```
XiaomikPlyuS, line 11: Duplicate of #

mikoMK: What do you mean by "Duplicate of #"?

PR #494, opened 2022-05-01
Fix variable name in variables marking guide

mikoMK:
**Marking guide of the variables skill test - task 3**
In "1." we are talking about the `myName` variable, but wrote `myAge`.


hamishwillee (approved):
Thank you - good catch!

PR #500, opened 2022-06-06
Update conditionals3-download.html

jarede-dev:
added the answers cause why not. lol


hamishwillee:
> added the answers cause why not

Well, perhaps because if I merged this then other users would no longer have a blank slate to work off?

Thanks. Closing.

PR #501, opened 2022-06-06
conditionals4-withAnswers

jarede-dev:
answered the conditional 4 cuz why not. lol


estelle (commented):
Please don't submit your solutions as PRs. We want others to be able to solve these learning area examples as well. Thanks.

PR #506, opened 2022-06-12
fix(readonly-confirmation): fix read-only prefix error

Tongky-HGU:
#### Summary

- fix CSS error in `:read-only` example

#### Motivation

- `:read-only` CSS is not working properly in Chorme.
- `-moz-` prefix won't work with Chome browser.
- Chome ignores whole CSS block with invalid prefix.

*As-Is*
image
*To-Be*
image

#### Supporting details

> A selector always goes together with a declaration block.
> When a user agent cannot parse the selector (i.e., it is not valid CSS 2.1),
> it must ignore the selector and the following declaration block (if any) as well.
> &mdash; [4.1.7 Rule sets, declaration blocks, and selectors](<https://www.w3.org/TR/CSS21/syndata.html#rule-sets:~:text=A%20selector%20always%20goes%20together%20with%20a%20declaration%20block.%20When%20a%20user%20agent%20cannot%20parse%20the%20selector%20(i.e.%2C%20it%20is%20not%20valid%20CSS%C2%A02.1)%2C%20it%20must%20ignore%20the%20selector%20and%20the%20following%20declaration%20block%20(if%20any)%20as%20well.>)

- [autoprefixer](https://autoprefixer.github.io/) result
```css
/* input */
input:read-only,
textarea:read-only {
  border: 0;
  box-shadow: none;
  background-color: white;
}
```

```css
/* output */
input:-moz-read-only, textarea:-moz-read-only {
  border: 0;
  box-shadow: none;
  background-color: white;
}
input:read-only,
textarea:read-only {
  border: 0;
  -webkit-box-shadow: none;
          box-shadow: none;
  background-color: white;
}
```

### DIFF
--- a/html/forms/pseudo-classes/readonly-confirmation.html
+++ b/html/forms/pseudo-classes/readonly-confirmation.html
@@ -61,16 +61,29 @@
         margin: 20px auto;
       }
 
-      input:-moz-read-only, textarea:-moz-read-only,
-      input:read-only, textarea:read-only {
+      input:-moz-read-only,
+      textarea:-moz-read-only {
         border: 0;
         box-shadow: none;
         background-color: white;
       }
 
-      textarea:-moz-read-write,
-      textarea:read-write {
+      input:read-only,
+      textarea:read-only {
+        border: 0;
+        -webkit-box-shadow: none;
+                box-shadow: none;
+        background-color: white;
+      }
+
+      textarea:-moz-read-write {
         box-shadow: inset 1px 1px 3px #ccc;
+        border-radius: 5px;
+      }
+
+      textarea:read-write {
+        -webkit-box-shadow: inset 1px 1px 3px #ccc;
+                box-shadow: inset 1px 1px 3px #ccc;
         border-radius: 5px;
       }
     </style>
--- a/html/forms/pseudo-classes/readonly-confirmation.html
+++ b/html/forms/pseudo-classes/readonly-confirmation.html
@@ -61,27 +61,14 @@
         margin: 20px auto;
       }
 
-      input:-moz-read-only,
-      textarea:-moz-read-only {
-        border: 0;
-        box-shadow: none;
-        background-color: white;
-      }
-
-      input:read-only,
-      textarea:read-only {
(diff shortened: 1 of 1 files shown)


teoli2003:
>     * `:read-only` CSS is not working properly in Chrome.

Can you elaborate on this? According to the browser compatibility data `:read-only` is supported since Chrome 1.

Also, Firefox supports `:read-only` without prefix for 2 years. 

I would prefer to remove all prefixes in this case.

html/forms/pseudo-classes/readonly-confirmation.html
```diff
@@ -61,18 +61,31 @@
...
         background-color: white;
       }
 
-      textarea:-moz-read-write,
-      textarea:read-write {
+      input:read-only,
+      textarea:read-only {
+        border: 0;
+        -webkit-box-shadow: none;
+                box-shadow: none;
```
teoli2003, line 75: Am I wrong in saying no browser support unprefixed `box-shadow`?

html/forms/pseudo-classes/readonly-confirmation.html
```diff
@@ -61,18 +61,31 @@
...
+      }
+
+      textarea:-moz-read-write {
         box-shadow: inset 1px 1px 3px #ccc;
         border-radius: 5px;
       }
+
+      textarea:read-write {
+        -webkit-box-shadow: inset 1px 1px 3px #ccc;
+                box-shadow: inset 1px 1px 3px #ccc;
```
teoli2003, line 86: Same here

teoli2003 (changes requested):

teoli2003:
@teoli2003, @wbamberg What is the policy with added an unprefixed CSS property when no browser supports it unprefixed (and all, including Gecko, with the `-webkit` prefix?

Tongky-HGU:
I think my PR is misleading. `:-moz-read-only` blocks whole css block include `:read-only`, so example is not working properly in Chorme.

I believe that all the above questions are discussed here. -> https://github.com/mdn/content/pull/17200

teoli2003 (approved):

vlakoff:
I've just added a comment on the sibling ticket: https://github.com/mdn/content/pull/17200#issuecomment-1888547101.

PR #530, opened 2022-08-04
SEMANTICS: add lang, viewport, alt, and role

estelle:


css/css-layout/practical-positioning-examples/fixed-info-box.html
```diff
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
-<html>
+<html lang="en-us">
 <head>
   <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width">
```
estelle, line 5:
```suggestion
  <meta name="viewport" content="width=device-width">
```

PR #535, opened 2022-08-18
Add missing title attribute on abbreviation

StartingCoding:
Of all of the abbreviation, this one was the only one without an expansion so I thought I would add it.


teoli2003:
Maybe we should replace it with BCE (Before Common Era)?

StartingCoding:
Nice, I didn't know about that type of notation.
I'm fairly new with PR on Github so I'm going to try to modify it and push it if it's ok.

lucalves (approved):

teoli2003 (approved):
👍

PR #540, opened 2022-09-01
remove unnecessary ARIA roles

scottaohara:
I'm currently making a PR for the wai-aria_basics file and in it there was reference to the markup of this file, where redundant ARIA roles were added to elements which had already been noted as exposing these semantics by default.  

As i'm removing these useless roles from the example in the referenced file, they should be removed from this example as well.


teoli2003 (approved):
Thank you!

PR #541, opened 2022-09-06
Made the loop more efficient

jalalbello:
⚡ put the least likely condition last


teoli2003:
Your change won't make any difference in speed (the same amount of tests).

PR #542, opened 2022-09-07
rename grid-gap to gap

tianheg:
https://developer.mozilla.org/en-US/docs/Web/CSS/gap


css/css-layout/grids/8-placement-starting-point.html
```diff
@@ -49,4 +49,4 @@ <h1>My article</h1>
 
 </body>
 
-</html>
\ No newline at end of file
+</html>
```
teoli2003, line 52:
```suggestion
</html>

```

teoli2003 (approved):
Thank you, I also run prettier to have it in the modern format.

PR #555, opened 2022-09-25
sync finished html files with examples in guides/javascript/buildingb…

TaebeomHeo:
Two things modified.

First, live output not syncing with code (especiall select tags) in Active learnings (A simple calendar and More color choices) could be solved with adding some inintializing.

Second, "More color choices" has mismatched selection value "psych" vs "psychedelic".

### DIFF
--- a/javascript/building-blocks/calendar-finished.html
+++ b/javascript/building-blocks/calendar-finished.html
@@ -76,6 +76,7 @@
         }
       }
 
+      select.value = 'January';
       createCalendar(31, 'January');
     </script>
   </body>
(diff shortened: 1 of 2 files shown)


dipikabh:
Note: This PR corresponds to https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/conditionals#active_learning_a_simple_calendar

dipikabh (approved):
Thank you @TaebeomHeo for your contribution and patience. These updates look good.

Making a note here that the initialization changes will need to be made to the code in the `mdn/content` repository as well.

PR #558, opened 2022-10-07
fix typo in index.js of video-store-offline example

TaebeomHeo:
fix typo in index.js located in video-store-offline example.


schalkneethling (approved):
Nice catch, @TaebeomHeo 🎉

PR #562, opened 2022-10-21
Update basic-text2-download.html

Souradeephazra123:
some words are converted into ordered list and some of them are into unordered list.


Souradeephazra123:
very good

Rumyra:
Hey @Souradeephazra123 - this pr is open on the main repo and thus I will close it to keep the example the same for others - work looks good tho!

PR #579, opened 2022-12-27
dl dt dd fixed for advanced-text1.html

ogzhncrt:
Missing HTML elements have been added to task 1
This PR fixes https://github.com/mdn/content/issues/23154

### DIFF
--- a/html/introduction-to-html/tasks/advanced-text/advanced-text1.html
+++ b/html/introduction-to-html/tasks/advanced-text/advanced-text1.html
@@ -34,12 +34,14 @@
 
       <h1>Advanced HTML Animals</h1>
 
-      Llama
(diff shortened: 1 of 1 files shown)


ogzhncrt:
Approved.

wbamberg:
Thanks for this PR!

However, this is not the right fix. You've updated the page that gets embedded at https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Test_your_skills:_Advanced_HTML_text#task_1 so it contains the solution, which defeats the point of this being a problem for the reader to solve.

The right fix I think is to include a new page that contains the solution, and link to that page separately. I've filed a PR for this at https://github.com/mdn/learning-area/pull/575.

PR #583, opened 2023-01-17
Added the missing Blush button

ayushitachi:
Blush button CSS and HTML are added


bsmth:
Thanks a lot for the suggestion! The button is already there: https://mdn.github.io/learning-area/css/css-layout/flexbox/complex-flexbox.html but it drops to the next line when the browser window is resized. I'm going to close this one for now. Thank you :)

PR #586, opened 2023-01-23
Update dom-example.html

Aether0x:


teoli2003:
I don't think this was intended to be changed here.

Issue #565, opened 2022-11-12
Order should be explained with ordinal numbers


innthomas:
Rather than use wink, smile, etc, using ordinal numbers like first, second, and third will make explaining easier with CSS order attribute.

Rumyra:
Hi @innthomas thanks for the issue - can you provide a link to where you see the issue - either MDN or in this repo - thanks!

innthomas:
Here is the link : https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox#horizontal_and_vertical_alignment

below is a screenshot 

![image](https://user-images.githubusercontent.com/34725225/202589272-31233449-2403-4c1a-ad01-e1110b1bba76.png)

Rumyra:
Thank you @innthomas 👍

dubaiandy:
不明觉厉

bsmth:
I've closed the linked PR because there is a bit of additional effort involved (screenshots and prose [on the content side](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox#nested_flex_boxes)). If you think that's a mistake, let me know and we can revisit this. Thank you :)

PR #587, opened 2023-01-24
Change order of the buttons to ordinal number #565

MAHIN0093:
Change order of the buttons to ordinal number


bsmth:
Thanks for the suggestions! If we do this one, we would have to update the screenshots and the text at https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox#nested_flex_boxes so I'm going to close this one for now. Let me know if you think that's a mistake.

Thank you :)

PR #588, opened 2023-01-26
Remove CSS box for HTML forms tasks

wbamberg:
In most of the "Test your skills" tasks for the HTML Forms module, the learner is only asked to provide HTML. But the editor also include an empty box for CSS, and this is confusing - because the box is empty, there's no indication that it is for CSS, and it might tempt people to try to add HTML in there.

For example: https://developer.mozilla.org/en-US/docs/Learn/Forms/Test_your_skills:_Other_controls#other_controls_1 .

This might have contributed to the confusion in https://github.com/mdn/content/issues/23776 (although that has other sources too).

So in this PR I've removed the CSS box from all editors under HTML/Forms except in the cases where the learner is actually expected to add CSS.

Maybe it would also be good to label the boxes?

### DIFF
--- a/html/forms/tasks/basic-controls/basic-controls1.html
+++ b/html/forms/tasks/basic-controls/basic-controls1.html
@@ -30,8 +30,6 @@
 
     </section>
 
-    <textarea class="playable playable-css" style="height: 130px;">
-
     </textarea>
 
     <textarea class="playable playable-html" style="height: 220px;">
--- a/html/forms/tasks/basic-controls/basic-controls2.html
+++ b/html/forms/tasks/basic-controls/basic-controls2.html
@@ -30,8 +30,6 @@
 
     </section>
 
-    <textarea class="playable playable-css" style="height: 130px;">
-
     </textarea>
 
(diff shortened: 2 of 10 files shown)


hamishwillee (approved):
Yes, this is way better.

> Maybe it would also be good to label the boxes?

Yes. Something like `<!-- CSS -->`?

html/forms/tasks/basic-controls/basic-controls1.html
```diff
@@ -30,8 +30,6 @@
 
     </section>
 
-    <textarea class="playable playable-css" style="height: 130px;">
-
     </textarea>
```
hamishwillee, line 33: @wbamberg Arrrgggg. Missed the end tags. Also, might be better to make as HTML comment because that way it is clear that you explicitly disabled the box rather than omitted it

hamishwillee: Fixed in https://github.com/mdn/learning-area/pull/589

PR #589, opened 2023-01-26
Fix dangling textarea closing tags

hamishwillee:
I merged #588 too quickly - noticed afterwards that the removal of the textarea elements did not include the end tag. This removes the end tags too.

FYI @wbamberg


wbamberg (approved):
👍 thank you @hamishwillee !

PR #591, opened 2023-02-06
"outline: transparent;" instead of "outline: none;"

AusDavo:
It may be preferable to specify a transparent outline. It preserves function for high contrast mode users. (Box shadow isn't applied in that mode.)

Some discussion of this principle here:
https://dev.to/eevajonnapanula/dont-set-outline-0-or-outline-none-for-focus-styles-2olb

### DIFF
--- a/html/forms/tasks/advanced-styling/marking.md
+++ b/html/forms/tasks/advanced-styling/marking.md
@@ -11,7 +11,7 @@ We've already provided you with a basic reset to build upon.
 1. First of all, try giving the search box a consistent width, height, padding, and border color across browsers. This is pretty self-explanatory; `line-height` tends to work better than `height` for keeping a consistent height in which the input text is vertically-centered. 
 2. You'll find that some browsers will not behave in terms of the form element's height. This is due to native OS styling being used in some cases. How can you remove this native styling? The answer here is `-webkit-appearance: none;`.
 3. Once you've removed the native styling, you'll need to add back one of the features it was providing, to keep same look and feel we originally had. How do you do this? Answer: rounded corners! We need `border-radius`.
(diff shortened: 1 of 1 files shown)


html/forms/tasks/advanced-styling/marking.md
```diff
@@ -11,7 +11,7 @@ We've already provided you with a basic reset to build upon.
 1. First of all, try giving the search box a consistent width, height, padding, and border color across browsers. This is pretty self-explanatory; `line-height` tends to work better than `height` for keeping a consistent height in which the input text is vertically-centered. 
 2. You'll find that some browsers will not behave in terms of the form element's height. This is due to native OS styling being used in some cases. How can you remove this native styling? The answer here is `-webkit-appearance: none;`.
 3. Once you've removed the native styling, you'll need to add back one of the features it was providing, to keep same look and feel we originally had. How do you do this? Answer: rounded corners! We need `border-radius`.
-4. One thing that is inconsistent across browsers (particularly looking at Safari here) is the position of the standard blue focus outline. How can you remove this? Answer: `outline: none;`
+4. One thing that is inconsistent across browsers (particularly looking at Safari here) is the position of the standard blue focus outline. How can you remove this? Answer: `outline: transparent;`
```
estelle, line 14: original was better.

html/forms/tasks/advanced-styling/marking.md
```diff
@@ -48,7 +48,7 @@ input[type="search"] {
   -webkit-appearance: none;
   border-radius: 6px;
 
-  outline: none;
+  outline: transparent;
```
estelle, line 51: The default value of `outline-style` is `none`. So, using the shorthand, setting `outline: transparent`, sets the outline-style to `none`. The original was better.

estelle (commented):
Thank you for your contribution. outline transparent is the same as outlne none, but the latter, the original, is better understood by developers.

PR #600, opened 2023-03-06
Make Live Sample keyboard accessible

dawei-wang:
Fix mdn/content#25017


wbamberg (approved):
👍 thank you for the fix @dawei-wang , and sorry to take so long to review it.

It works great, except I couldn't see the focus ring with the colors chosen, so I have changed them. According to the Firefox accessibility tool, contrast is still good with these colors.

PR #607, opened 2023-03-28
Update enabled-disabled-shipping.html

CalinObo:
Took the border off of the check button.


bsmth:
I think this one is okay as-is so I'm going to close it for now. If you think I've made a mistake, please let me know and we can revisit it. Thank you :)

PR #614, opened 2023-05-01
mdn objects

Veronicandemo:


bsmth:
We're not merging in changes from this as it would fix the template that other readers should use as a starting point. Thank you :)

PR #616, opened 2023-05-04
Modify random number generating function and indentation

harshvsri:
Previously, the random function was generating only single output ( as 1) .
Now, it will generate integers ranging from [1, 100].
github-correction

### DIFF
--- a/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
+++ b/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en-us">
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
 
     <title>Number guessing game</title>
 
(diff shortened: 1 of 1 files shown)


javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
```diff
@@ -23,94 +23,96 @@
...
+      guessField.value = "";
+      guessField.focus();
 
-	  lastResult.style.backgroundColor = 'white';
+      lastResult.style.backgroundColor = "white";
 
-	  randomNumber = Math.floor(Math.random()) + 1;
-  }
-</script>
+      randomNumber = Math.floor(Math.random() * 100) + 1;
```
harshvsri, line 115: Well this the prime change made

bsmth:
Thanks a lot for this one @harshvsri - the formatting changes look good, but we should leave the random number generator as-is for now as the improvements are supposed to be found while following along with the article. See https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_went_wrong#a_logic_error

I'm going to close for now, but thanks for the submission :)

PR #617, opened 2023-05-07
rm x-ua-compatible

wbamberg:
Fixes https://github.com/mdn/content/issues/26623.

According to [our own docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv):

> x-ua-compatible: If specified, the content attribute must have the value "IE=edge". User agents are required to ignore this pragma.  

Some of the usages here were incorrect and giving validator errors. I think it's best to remove it everywhere.

This PR is in 2 commits:

* https://github.com/mdn/learning-area/commit/adde4d181696908c978d57f4756a9bf0efeb109c: Prettier noise
* https://github.com/mdn/learning-area/commit/c1a0cb863cee00d36ec043a13ddc6ef512ff804c: substantive changes

@teoli2003 , can you review?

### DIFF
--- a/javascript/apis/drawing-graphics/threejs-cube/index.html
+++ b/javascript/apis/drawing-graphics/threejs-cube/index.html
@@ -2,7 +2,6 @@
 <html lang="en-us">
   <head>
     <meta charset="utf-8" />
-    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
     <title>Three.js basic cube example</title>
--- a/javascript/apis/drawing-graphics/threejs-cube/index.html
+++ b/javascript/apis/drawing-graphics/threejs-cube/index.html
@@ -1,17 +1,16 @@
 <!DOCTYPE html>
 <html lang="en-us">
   <head>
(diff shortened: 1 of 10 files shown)


Josh-Cena:
O no... If we do care about validators (or at least not letting the validator confuse learners) the trailing slashes here are surely going to be problematic.

wbamberg:
> O no... If we do care about validators (or at least not letting the validator confuse learners) the trailing slashes here are surely going to be problematic.

If you're talking about the trailing slashes Prettier adds, then:
- we've already had this conversation and decided apparently that we will accept Prettier here. I don't personally agree but don't see why we want to have this conversation again.
- the validation generates an info message for the trailing slashes, not an error

Josh-Cena:
Thanks. In general I've been saddened by Prettier's behavior on HTML, but it seems okay here.

teoli2003 (approved):
👍

teoli2003:
I'm happy to see `X-UA-Compatible` go. I always hated it (and there are not many things I hate). 

For Prettier, I love it with JS, mostly like it with CSS (there are a few oddities), and don't like it for HTML. One day we'll find an alternative (a Prettier-for-docs ?): it is not irreversible.

PR #619, opened 2023-05-13
Fix the integer square number loop

Disney-Banje:
Hey there, just wanted to give you a heads-up that I made a quick update to the program. Now, it'll only accept valid numbers, which makes it more accurate and easier to read. Let me know what you think about the change!

### DIFF
--- a/javascript/building-blocks/loops/integer-squares.html
+++ b/javascript/building-blocks/loops/integer-squares.html
@@ -24,7 +24,7 @@
 
     btn.addEventListener('click', () => {
       para.textContent = 'Output: ';
-      const num = input.value;
+      const num = Number(input.value);
       input.value = '';
       input.focus();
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thank you. It is an improvement to enforce a number, but there are better ways to do this. `Number` represents tells us this is a real number when we want an integer here. So it is kind of half way, and there is no fall back handling if a non-number is entered.

A better way to do this would be to change the input type from text to number and put a step in there: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

In addition, any changes here, would also have to be mirrored in the associated content page: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Looping_code

I'm going to close this, but would welcome an update in line with this suggestion. Thanks very much for taking the time time to post and help all users of this content out.

PR #625, opened 2023-05-28
Replaced ' with " in one of the code examples

arka1002:
It won't be a single quote because there's a single quote in "Don't"


bsmth:
Hi @arka1002 - thanks a lot for the suggestion. Your fix looks like a good start, but we won't merge your pull request because doing so would fix the problem so others wouldn't be able to try for themselves. Instead you can go here: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Test_your_skills:_Strings#strings_1 and use the embedded editor to try some solutions.

I'm going to close this one for now. Thank you and well done :)

PR #626, opened 2023-05-29
changes made

advitiya7:
i have made the changes


bsmth:
Thanks for opening this @advitiya7 - I'm going to close this because we're editing in the wrong place, we shouldn't get rid of the `<abbr>` elements which are useful for illustration.

PR #627, opened 2023-05-30
fix: update the URLs to mdn social logo

yin1999:
update the URLs to mdn social logo, this is a part of reflecting changes to mdn/content#27046.


caugner (approved):

bsmth (approved):
Thanks @yin1999 and @caugner 👍🏻

PR #628, opened 2023-06-02
[revise] Update payment-form.html

CBID2:
### Description
I added a `required="true"` attribute in the code snippet.

### Motivation

According to the [W3C](https://www.w3.org/TR/WCAG20-TECHS/ARIA2.html), it is best to add the 'aria-required' property with the value of "true" if users are required to fill out certain sections of a form.

### Related issues and pull requests

Relates to Issues [#26790](https://github.com/mdn/content/issues/26790) and Issue [#629](https://github.com/mdn/learning-area/issues/629). It also relates to pull request [#27035](https://github.com/mdn/content/pull/27035).

### DIFF
--- a/html/forms/html-form-structure/payment-form.html
+++ b/html/forms/html-form-structure/payment-form.html
@@ -48,14 +48,14 @@
                 <span>E-mail: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="email" id="mail" name="usermail">
+              <input type="email" id="mail" name="usermail" required="true">
             </p>
             <p>
               <label for="pwd">
                 <span>Password: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="password" id="pwd" name="password">
+              <input type="password" id="pwd" name="password" required="true">
             </p>
         </section>
         <section>
@@ -75,7 +75,7 @@
                 <span>Card number:</span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-                <input type="tel" id="number" name="cardnumber">
+                <input type="tel" id="number" name="cardnumber" required="true">
             </p>
             <p>
               <label for="expiration">
--- a/html/forms/html-form-structure/payment-form.html
+++ b/html/forms/html-form-structure/payment-form.html
@@ -41,7 +41,7 @@
                 <span>Name: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="text" id="name" name="username">
+              <input type="text" id="name" name="username" required="true">
             </p>
             <p>
               <label for="mail">
--- a/html/forms/html-form-structure/payment-form.html
+++ b/html/forms/html-form-structure/payment-form.html
@@ -41,21 +41,21 @@
                 <span>Name: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="text" id="name" name="username" required="true">
+              <input type="text" id="name" name="username" required>
             </p>
             <p>
               <label for="mail">
                 <span>E-mail: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="email" id="mail" name="usermail" required="true">
+              <input type="email" id="mail" name="usermail" required>
             </p>
             <p>
               <label for="pwd">
                 <span>Password: </span>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Looks good, though would be good to have all fixes in the same PR.

How come the name and expiration date are not required? Do you think they shouldn't be?

CBID2:
> Looks good, though would be good to have all fixes in the same PR.
> 
> How come the name and expiration date are not required? Do you think they shouldn't be?

Hi @hamishwillee. I considered adding it to my other PR but I felt that it would possibility cause more problems, so I kept them separate. As far as the name and expiration date, good point. I'll make those changes

html/forms/html-form-structure/payment-form.html
```diff
@@ -41,21 +41,21 @@ <h2>Contact information</h2>
                 <span>Name: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="text" id="name" name="username">
+              <input type="text" id="name" name="username" required="true">
```
hamishwillee, line 44:
```suggestion
              <input type="text" id="name" name="username" required>
```

html/forms/html-form-structure/payment-form.html
```diff
@@ -41,21 +41,21 @@ <h2>Contact information</h2>
...
             </p>
             <p>
               <label for="mail">
                 <span>E-mail: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="email" id="mail" name="usermail">
+              <input type="email" id="mail" name="usermail" required="true">
```
hamishwillee, line 51:
```suggestion
              <input type="email" id="mail" name="usermail" required>
```

html/forms/html-form-structure/payment-form.html
```diff
@@ -41,21 +41,21 @@ <h2>Contact information</h2>
...
             </p>
             <p>
               <label for="pwd">
                 <span>Password: </span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-              <input type="password" id="pwd" name="password">
+              <input type="password" id="pwd" name="password" required="true">
```
hamishwillee, line 58:
```suggestion
              <input type="password" id="pwd" name="password" required>
```

html/forms/html-form-structure/payment-form.html
```diff
@@ -75,7 +75,7 @@ <h2>Payment information</h2>
                 <span>Card number:</span>
                 <strong><span aria-label="required">*</span></strong>
               </label>
-                <input type="tel" id="number" name="cardnumber">
+                <input type="tel" id="number" name="cardnumber" required="true">
```
hamishwillee, line 78:
```suggestion
                <input type="tel" id="number" name="cardnumber" required>
```

html/forms/html-form-structure/payment-form.html
```diff
@@ -41,21 +41,21 @@ <h2>Contact information</h2>
                 <span>Name: </span>
                 <strong><span aria-label="required">*</span></strong>
```
hamishwillee, line 42: On all of these, should it be `aria-label="required"` or `aria-required="true"`? I suspect the former but will check.

hamishwillee: The answer I got is that this is good.

> the aria-required=true would be on the form control. it isn’t necessary if the required attribute is set and valid, but can be used on a component or element that is required but isn’t a form control.

So we don't need it because we're setting required on the fields.

> The aria-label="required"  is appropriate for the *, as it labels the asterisk, providing an appropriate accessible name to the character that means “required”  but wouldn’t otherwise mean that.
Best practices are to include a visible comment for ALL users explaining what the asterisk means.

CBID2: Ok

hamishwillee:
The associated PR still needs some work. This can merge when that is correct.

CBID2:
> The associated PR still needs some work. This can merge when that is correct.

What's missing @hamishwillee?

hamishwillee:
@CBID2 No. Your other PR seemed pretty much unrelated to this one, so while I was able to kill the errors, it didn't "make sense".

I have created https://github.com/mdn/content/pull/27395 which matches this PR. As soon as someone in the team review that we can merge both of these.

PR #633, opened 2023-06-15
chore(html): use const, format code example

bsmth:
Fixes https://github.com/mdn/content/issues/25750 :

>I expected the github link to display the same code that is in the snippet below the code, or vice versa. They should match.

See https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Advanced_text_formatting#representing_computer_code


bsmth:
Very good. I will follow up in content with a clarification it's not the entire file. Thank you!

PR #647, opened 2023-07-04
Update learning-area/css/styling-boxes/cool-information-box-start /index.html

devusexu:
The paragraph element should be a div element according to the doc https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/A_cool_looking_box#:~:text=style%20the%20provided-,%3Cdiv%3E,-%2C%20giving%20it%20the


bsmth:
Thanks a lot, it looks like this change has already been applied, so I'm closing this one for now. Thank you :)

PR #648, opened 2023-07-07
Add brackets to match webpage in JavaScript -> Object building practice

fransan6:
On the [web page](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_building_practice#getting_started), there are brackets so I have added brackets to the downloadable js file. Whether the webpage or downloadable file is updated, I believe they should be consistent.

Screenshot 2023-07-07 211251


bsmth (approved):
Thank you! 🎉

PR #651, opened 2023-07-16
fixed typeError

bampots97:
number-game-errors.html:86


bsmth:
Hi there, the changes you've submitted are correct, well done.

Issue #642, opened 2023-05-23
Error Message is Cut Off at the Top when the Example is at the Bottom of the Browser Window


jfc3-dol:
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation

### What specific section or headline is this issue about?

Constraining the values of your entries

### What information was incorrect, unhelpful, or incomplete?

The error message is cut off at the top when the example is at the bottom of the browser window. A reminder that you might have similar issues with the bottom of the error message.

Please see the attached screenshot.

<img width="960" alt="mdn - The error message is cut off at the top when the example is at the bottom of the browser window Using Chrome on Windows 10" src="https://github.com/mdn/content/assets/10423487/5b2e7811-667e-419d-879e-035d858d7eb4">

**NOTE** - I’m using Chrome on Windows 10.

### What did you expect to see?

I expected to see the entire error message and top only part of it.

### Do you have any supporting links, references, or citations?

nope

### Do you have anything more you want to share?

nope

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/learn/forms/form_validation`
* MDN URL: https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/learn/forms/form_validation/index.md
* Last commit: https://github.com/mdn/content/commit/d85a7ba8cca98c2f6cf67a0c44f0ffd467532f20
* Document last modified: 2023-04-05T04:57:48.000Z

</details>

bsmth:
Thanks for raising this one @jfc3-dol - the example is here: https://github.com/mdn/learning-area/blob/main/html/forms/form-validation/fruit-length.html

If you would like to have a go at fixing this through additional top padding or margin. It looks like this is Chrome-specific behavior as the hint is displayed as expected on Safari and Fx

jfc3-dol:
@bsmth 

Someone else can do it since I was only pointing out the issue. Plus, I am not sure of the process and can't do so on my work computer.

harshvsri:
I would love to fix it...

Light13008:
please could you assign this issue to me ?
I'd love to fix the issue.

hamishwillee:
We no longer assign issues, but rather if you want to handle an issue, start a PR and cross link as `Fixes #XXXX`. Then other people who are interested should first check that no other PRs are in progress.

vanshdhar999:
[...]

PR #653, opened 2023-08-03
#642 Fixed margin of the form element

vanshdhar999:
Added margin-top style to form element to show the full invalid text after clicking submit button. Works in Chrome browser now. 
Screenshot 2023-08-03 at 1 52 01 PM


bsmth:
Thanks a lot for opening this one. It's been fixed in https://github.com/mdn/learning-area/pull/656 already.

PR #656, opened 2023-08-07
fix(forms): add margin to form for readability

bsmth:
The validity error message has a low z-index on Chrome, so it's occluded by the rest of the page in an iframe. This makes the example at https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#constraining_the_values_of_your_entries have a semi-occluded validity message.

This PR adds some padding so that the message is visible when constraint validation fails.

<details>
<summary>Screenshot:</summary>

image

</details>


bsmth:
Thank you

PR #670, opened 2023-09-27

Karn2003:


Karn2003:
j

bsmth:
Hi there, the solution you're suggesting looks great, well done 🎉

We're not merging in changes for this as it would fix the problem that other readers should try to find. Thank you :)

Issue #609, opened 2023-04-11
html/forms/html-form-structure/payment-form.html STYLING ISSUE


vvronskyFX:
Inside the MDN Learning area [article](https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form), The iframe source links to https://live-samples.mdn.mozilla.net/en-US/docs/Learn/Forms/How_to_structure_a_web_form/_sample_.a_payment_form.html that seems to be from an old version of the page.

Rather than looking like this:
![image](https://user-images.githubusercontent.com/64465297/231058588-14c95728-8bd5-43c6-bacf-b992374565c0.png)

The page looks like this:
![image](https://user-images.githubusercontent.com/64465297/231058661-198edcc3-ba02-44e3-8e9d-90ffc6c04c7d.png)

Note the spacing differences in both screenshots. 

Here's what I see in the browser inspect. The links to the source and the iframe are different. 
![Screenshot 2023-04-11 at 12 50 39 AM](https://user-images.githubusercontent.com/64465297/231059616-8962e287-e607-4b8b-a3b3-f8513d39e79c.png)

habeebmoosa:
I want to work on this issue. So could you assign it to me?

vvronskyFX:
> 

Hey @habeebmoosa I'm not sure how to assign this.

marlatte:
I noticed the same thing. The problem is in the `payment-form.css` file, where `label span` (line 23) applies the styling to the label's contents instead of the whole label. I got it to look nicer by changing this the rule to `p label`, though I'm sure there's more than one way to do it.

PR #678, opened 2023-10-26
Remove two files to reduce sync work between content and learning-area - requires content PR

zfox23:
PR 2/2 to resolve https://github.com/mdn/learning-area/issues/609.

The other PR is here: https://github.com/mdn/content/pull/29838


bsmth:
This makes sense as we have the code:

1. In this repo
2. On [the article itself](https://github.com/mdn/content/blob/main/files/en-us/learn/forms/how_to_structure_a_web_form/index.md?plain=1)
3. In an [example page](https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form/example) that's a child of the article

I don't see any links to these after your content PR is merged, but it might be good to check elsewhere. If we merge, do you want to [reply here](https://discourse.mozilla.org/t/payment-form-structure-help-vv/112773) that the final code can be found at https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form/example?

bsmth (approved):
I think we're good to remove these now given [the rationale](https://github.com/mdn/learning-area/pull/678#issuecomment-1782742453) and the recent improvements on the content side.

Thanks, Zach 🙌🏻

PR #694, opened 2023-12-10
fix(html): update srcset-resolutions demo

OnkarRuikar:
- related to https://github.com/mdn/content/issues/3397

The PR
- adds script to show actual loaded image name on screen
- runs Prettier on the file

On touch devices developer tools are not available. The change makes easy to realize which image has been fetched and loaded by the browser.


hamishwillee (approved):
This is better.

PR #709, opened 2024-01-16
Update text-start.html archivos de recuperación de facebook 

Martyn2747:
Espero y les sirva


bsmth:
I'm going to close this one for now as this change doesn't look like it's in the right place. Additionally, we're not changing this file as it would fix the starting template that other readers should use. 
Thank you :)

PR #711, opened 2024-01-18
Wrap code in `<code>` tags

qadzek:
This enhances the visual distinction between regular text and code.


bsmth (approved):
Thanks a lot! I think it would be great to have all of this in markdown (instead of HTML inside markdown) eventually, but this is a nice improvement!

PR #712, opened 2024-01-21
refactor: update js to match webpage

fransan6:
The JavaScript in some files does not match the code blocks online and an earlier JavaScript file. I have refactored them for consistency to match the main [webpage](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_building_practice) on this, with Prettier adjusting indentation and quotation marks.


bsmth (approved):
Brilliant, thanks a lot!

PR #718, opened 2024-02-08
update formatting of the li button to match the solution

wjdittmar:
changed the size of the li button font to match the solution


bsmth (approved):
Looks good to me, thanks!

PR #721, opened 2024-02-19
Update links1-download.html

Fanowell121:


bsmth:
Hi there! We're not merging in changes from this as it would fix the starting template that other readers should use. Thank you :)

PR #724, opened 2024-03-19
Fix too permissive regular expression

qadzek:
The dot needs to be escaped due to its special meaning in regular expressions. 

Currently, both `j.doe` and `jXdoe` are considered valid, contrary to the task requirement that specifies that a single letter should be followed by a dot.

I made this PR before (#710) but I accidentally closed that one.


bsmth (approved):
Super, thanks a lot for the improvements! Merging this shortly 👍🏻

PR #725, opened 2024-03-23
Update form validation phone solution to work with v flag

fabulousgk:
The current version does not work, this is because of the change to using the v flag for regex requiring - to be escaped.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern


bsmth (approved):
Well spotted! Thanks a lot for your contribution!

PR #727, opened 2024-04-02
Update style-resources.txt

prakhar5100:
wrong syntax of background-color


bsmth:
Hi there, thanks a lot for suggesting the changes. You're absolutely right, but I'm not going to merge this because the typo is an intentional mistake that readers should discover and fix themselves:

>Beware! There are two errors in the provided rulesets. Using any technique that you know, track these down and fix them before moving on.

I'm going to close this now, but thank you and well done on fixing it :)

PR #731, opened 2024-04-16
fix: typo in showing cat's name

TheKostyak:
Showing ReferenceError: catName is not defined. Use istead cat.name


pepelsbey (approved):
Good catch! Thank you :)

PR #735, opened 2024-05-24
Update number-guessing-game.html

bruke3:
makes it look nice. thats all


pepelsbey:
Thank you for the suggestion! However, we’d rather keep a simpler design using built-in browser styles to make the demo less distracting.

PR #743, opened 2024-06-25
Reintroduced intentional referenceError for tutorial

simonas-kulakauskas:
Fixed someone's previous commit, this was **meant** to return a referenceError that was then to be fixed by the tutorial taker by creating a variable 'catError' and assigning it the cat.name value.

This same line can be seen in the live preview found at [Live Preview HTML](https://github.com/mdn/learning-area/blob/main/javascript/oojs/tasks/object-basics/object-basics1.html). 

Tutorial 'Test your skills: Object basics - Object Basics 1' does not make sense without it.

### DIFF
--- a/javascript/oojs/tasks/object-basics/object-basics1-download.html
+++ b/javascript/oojs/tasks/object-basics/object-basics1-download.html
@@ -46,7 +46,7 @@
     let para1 = document.createElement('p');
     let para2 = document.createElement('p');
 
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Super, thank you. The marking guide for this is here: https://github.com/mdn/learning-area/blob/main/javascript/oojs/tasks/object-basics/marking.md

PR #744, opened 2024-06-26
Update letter-text.txt

Alenjoe235:


dipikabh:
Hello @Alenjoe235, as this exercise states https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Marking_up_a_letter#starting_point, `letter.txt` is just the raw text file. You need to download it, mark it up in, and save as an `.html` file. This text file itself does not not need to be marked up.

PR #745, opened 2024-06-29
fix(css): use buttons instead of links and checkboxes in positioning examples

OnkarRuikar:
Part of the fix for:
- https://github.com/mdn/content/issues/24806
- https://github.com/mdn/content/issues/30473

The PR:
- uses buttons instead of links and checkboxes. The examples could use the `tab` and `space`/`enter` buttons.
- adds ARIA attributes

To test, open the HTML files in a browser.

I'll submit content PR once this is reviewed and finalized.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,3 @@
+{
+  "bracketSameLine": true
+}
--- a/css/css-layout/practical-positioning-examples/fixed-info-box.html
+++ b/css/css-layout/practical-positioning-examples/fixed-info-box.html
@@ -37,15 +37,12 @@
       }
 
       .info-box [role="tab"] {
-        display: inline-block;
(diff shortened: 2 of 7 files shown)


pepelsbey (commented):
That’s a big improvement already, but I wonder if we can make it even better. The current script doesn’t manage the focus for switching tabs, which is an important part of the accessible experience. Since this demo is focused on the layout, we can put [the script from APG](https://www.w3.org/WAI/content-assets/wai-aria-practices/patterns/tabs/examples/js/tabs-manual.js) right next to the index.html and make it work its magic.

This script relies only on attributes and doesn’t require any class names. It will keep the demo even simpler and more focused.

pepelsbey (approved):
Looks great :) Thank you

PR #747, opened 2024-07-02
Update shopping-list-finished.html

SukhbirSinghKhalsa:
Added the conditional statement to check input is not empty and trim to remove extra spaces.
This ensures that user cannot add empty items in list.
We can also add it to the documentation example as step 12.

Example link:
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Manipulating_documents

let me know if there are any questions around it.

### DIFF
--- a/javascript/apis/document-manipulation/shopping-list-finished.html
+++ b/javascript/apis/document-manipulation/shopping-list-finished.html
@@ -1,44 +1,46 @@
 <!DOCTYPE html>
 <html lang="en-US">
-  <head>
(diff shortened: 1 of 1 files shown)


pepelsbey:
Hey! Thank you for the suggestion. It makes perfect sense in production code. However, it might be too much for the “Manipulating DOM” chapter. We’re always trying to stick to the point and not overwhelm readers. Let’s keep it focused :)

PR #749, opened 2024-07-11
Update random-color-addeventlistener.html

Caroylina:


bsmth:
Thanks a lot for opening this one, it looks like you have the right idea! We're not merging in changes from this as it would fix the starting template that other readers should use. Thank you :)

PR #753, opened 2024-07-21
Update images1-download.html

Tichman92:
code updated


estelle:
This is the base code for everyone to edit, so we don't want to change the example file. Closing.

PR #755, opened 2024-07-31
Update flex-align0.html

AshuSharma199:
change in border


estelle:
Hi @AshuSharma199 

Thank you for the suggested edit. I think the example looks better with a single-pixel border. I was wondering if there is a reason to increase the border to two-pixels. Thanks for clarifying for me.

estelle:
Closing for now as we await a response. Please re-open if you still want this addressed. Thanks.

PR #756, opened 2024-08-01

jojotjo:
correct the error event to Event 86 line


bsmth:
Well done for fixing. I'm not going to merge this change because it's an intentional error that should be left in for others to find. See https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_went_wrong#fixing_syntax_errors. Thank you :)

PR #759, opened 2024-08-13
Update links2-download.html

Hugibald:


pepelsbey:
Hey! The examples in the learning area should remain incomplete so everyone can play with them on their own.

PR #762, opened 2024-09-04
Remove trailing whitespaces from meta-example.html

otakutyrant:


pepelsbey (approved):
Good catch! Thank you :)

PR #763, opened 2024-09-05
Update number-guessing-game-start.html

elinolasco:
Adding functions.


bsmth:
We're not merging in changes from this as it would change the starting template that other readers should use. Thank you :)

PR #764, opened 2024-09-07
fix: incorrect error example

RasmusKard:
### Description

Changed const to let

### Motivation

Currently the "other common errors" section's subsection "The program always says you've won, regardless of the guess you enter" is incorrect as instead of the described behavior it throws an error. (TypeError: invalid assignment to const 'userGuess')

(It might be better to keep this as const and change docs instead to reflect the current behavior, but I'm not experienced enough to feel comfortable writing docs)

### Additional details

Docs related to this file:

https://github.com/mdn/content/blob/b67fd42cfb01dd4d9504c4182b462851588a0bad/files/en-us/learn/javascript/first_steps/what_went_wrong/index.md#L4

### DIFF
--- a/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
+++ b/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
@@ -55,7 +55,7 @@
 
   function checkGuess() {
 
-    const userGuess = Number(guessField.value);
+    let userGuess = Number(guessField.value);
     if(guessCount === 1) {
       guesses.textContent = 'Previous guesses: ';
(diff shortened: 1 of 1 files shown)


bsmth:
Thanks for opening. The "other common errors" section is generic https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_went_wrong#other_common_errors

It's talking about "common errors you'll come across in your code", and isn't specific to this example, so we can keep the code as-is for now. I'll close this PR for the moment so the code is aligned with the content. Thank you!

PR #766, opened 2024-09-12
refactor: move the days determination logic to the createCalendar function

### LINKED ISSUE #469: Another way to write the calendar (关于日历的另一种写法)
按照月份1、3、5、10、12为31天，2月不闰月28天，其余时间30天。这样借助三元运算符就不用在`createCalendar`中传入`days`。
那么这种写法和历程中给出的传days的写法有什么区别呢？ 即使考虑闰月，对于两种写法都是加入判断语句，似乎与不闰月没有不同。

> According to the 1st, 3rd, 5th, 10th and 12th of the month, there are 31 days. February is not a leap month with 28 days, and the rest of the month has 30 days. In this way, with the help of the ternary operator, there is no need to pass in days in createCalendar.
So what is the difference between this way of writing and the way of writing days given in the process? Even if leap months are taken into account, judgment statements are added to both writing methods, which seems to be no different from non-leap months.

```js
var select = document.querySelector('select');
var list = document.querySelector('ul');
var h1 = document.querySelector('h1');

select.onchange = function() {
  var choice = select.value;

  createCalendar(choice);
}

function createCalendar(choice) {
  (choice === 'February')? days = 28
  (choice === 'January'||choice ==='March'||choice ==='May'||choice ==='July'||choice ==='August'||choice ==='October'||choice ==='December')? days=31:days=30;
  list.innerHTML = '';
  h1.textContent = choice;
  for (var i = 1; i <= days; i++) {
    var listItem = document.createElement('li');
    listItem.textContent = i;
    list.appendChild(listItem);
  }
}

createCalendar(choice);
```

yin1999:
Fixes: #469

content PR: mdn/content#35850

### DIFF
--- a/javascript/building-blocks/calendar-finished.html
+++ b/javascript/building-blocks/calendar-finished.html
@@ -49,35 +49,40 @@
     </ul>
 
     <script>
-      const select = document.querySelector('select');
-      const list = document.querySelector('ul');
-      const h1 = document.querySelector('h1');
+      const select = document.querySelector("select");
+      const list = document.querySelector("ul");
+      const h1 = document.querySelector("h1");
 
-      select.onchange = function() {
+      select.addEventListener("change", () => {
         const choice = select.value;
+        createCalendar(choice);
+      });
+
+      function createCalendar(month) {
         let days = 31;
 
-        if(choice === 'February') {
+        if (month === "February") {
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Super, thank you! 👍🏻

PR #767, opened 2024-10-05
Display #000 colour input box highlight on forcus

josuke0227:
This commit is related to the following:
[https://github.com/mdn/content/pull/36103](https://github.com/mdn/content/pull/36103/commits)

Since the outline colour in which the input elements are highlighted differs from operation systems - blue by default on Mac OS, black (rgb(16, 16, 16)) on WIndows - this commit is to apply the same outline colour across those operation systems.

This implementation was tested on Chrome and Firefox.


hamishwillee (approved):
Thank you very much!

PR #770, opened 2024-10-08
Button text change

Chaya90:


javascript/building-blocks/tasks/events/events1-download.html
```diff
@@ -30,6 +30,13 @@
 
   <script>
     const btn = document.querySelector('.off');
+    const newText = btn ? "Machine is off" : "Machine is on"
+    const clickButton = function (newText) => {
```
zeenox-stack, line 34: use the normal function rather than using the arrow syntax here.

javascript/building-blocks/tasks/events/events1-download.html
```diff
@@ -30,6 +30,13 @@
...
+      clickButton.addEventListener("onclick", (newText) => {
+        console.log (newText)
+        
+      })
+    })
```
zeenox-stack, line 40: You also have to call the clickButton variable with function with the newText as its argument to make sure it is working properly. Something like this:

clickButton(newText);

zeenox-stack (changes requested):
The comments should help you solve the problems

pepelsbey:
Hey! This is a sample file for anyone to use as a [starting point](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Test_your_skills:_Events#events_1) in testing their skills. Feel free to practice in your fork :)

PR #772, opened 2024-10-22
Replaced onchange to EventListener to detect change

bhagyeshsp:
The MDN example that redirects to this code uses EventListener method to detect change in select element. I learned that standard practice is using EventListener method, hence the change.


bsmth (approved):
Super, thanks 👍🏻

PR #773, opened 2024-10-22
Replace onchange with EventListener method

bhagyeshsp:
The MDN example that redirects to this code uses EventListener method to detect change in select element. I learned that standard practice is using EventListener method, hence the change.

### DIFF
--- a/javascript/building-blocks/simple-else-if.html
+++ b/javascript/building-blocks/simple-else-if.html
@@ -21,7 +21,7 @@
       const select = document.querySelector('select');
       const para = document.querySelector('p');
 
-      select.onchange = setWeather;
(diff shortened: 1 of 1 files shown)


bsmth:
Content equivalent: 

https://github.com/mdn/content/blob/fd9a16a3a56e295eb80fc24a5dbd56a842b2958e/files/en-us/learn/javascript/building_blocks/conditionals/index.md?plain=1#L127

bsmth (approved):
Nice, thank you and well done on your first contrib here 🎉

bhagyeshsp:
Thank you so much @bsmth for approving and validating my understanding!

PR #775, opened 2024-10-28
Number Game Errors: Move script inside body tag

mao-sz:
## Because

In the ["What went wrong?" exercise](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/What_went_wrong), there is a warning about line numbers in error messages potentially not matching source code line numbers due to code injection from extensions like live-server.

This is a result of the code's script tag being placed *after* the closing body tag, so live-server's injected code comes before it.

By placing the script tag *before* the closing body tag, live-server injects code after it, preserving line numbers in the error messages, making the lesson much easier to follow for more people. The relevant warning can then be removed from the lesson.

This has been tested with [VSCode's Live Server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer), [VSCode's Live Preview extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.live-server) and the [live-server npm package](https://www.npmjs.com/package/live-server).

## This PR

- Moves the closing body tag down *after* the script tag closes.
- Fixes inconsistent indentation across the file to only use 2 spaces per level (rather than mixing spaces and tabs).
- Uses appropriate indentation for each block as necessary.

## Additional information

This PR depends on (https://github.com/mdn/content/pull/36541) which amends the line number references and screenshots in the lesson itself to match what these code changes will produce.

### DIFF
--- a/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
+++ b/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
@@ -23,94 +23,92 @@
   </head>
 
   <body>
-      <h1>Number guessing game</h1>
+    <h1>Number guessing game</h1>
 
-      <p>We have selected a random number between 1 and 100. See if you can guess it in 10 turns or less. We'll tell you if your guess was too high or too low.</p>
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 yes, works for me with VSCode as well (and removes an error from the W3C validator).

PR #776, opened 2024-10-31
Number Game Errors: Format with Prettier

mao-sz:
## Because

As per (https://github.com/mdn/content/pull/36541#issuecomment-2450332532), the MDN code standard uses Prettier. This repo hasn't fully caught up but this exercise can be formatted as the changes in that PR would reference this code's line numbers. Using Prettier here would also mean fewer discrepancies in case any learners happen to already have Prettier configured (e.g. formatting on autosave) - not too unreasonable of a situation.

## This PR

- Formats the number game html file with Prettier's default rules.

### DIFF
--- a/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
+++ b/javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
@@ -1,7 +1,7 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en-US">
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
 
     <title>Number guessing game</title>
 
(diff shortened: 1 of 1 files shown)


MaoShizhong:
@wbamberg Formatting changes made. I'll get the other PR amended in due course.

javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
```diff
@@ -1,7 +1,7 @@
-<!DOCTYPE html>
+<!doctype html>
```
wbamberg, line 1: For some reason my Prettier uppercases `doctype`, but it doesn't affect line or character numbers, so no matter.

wbamberg (approved):
👍 thank you @MaoShizhong !

javascript/introduction-to-js-1/troubleshooting/number-game-errors.html
MaoShizhong: Huh, interesting. The VSCode extension does indeed capitalise it whereas running Prettier through a Neovim plugin or via their CLI (v3.3.3) is happy with it lowercase. But yeah, no matter since it doesn't cause any meaningful discrepancies, but odd nonetheless.

PR #785, opened 2025-01-09
Final Code Preview Different from Code Along Example #784

danieletie:
This is a PR to update the live code and code preview to match the code along examples in this [lesson](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Your_first_form)


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #787, opened 2025-03-09
Changes for issue #38530 on mdn/content

low-perry:
Issue [#38530](https://github.com/mdn/content/issues/38530) on mdn/content could be solved by these changes.
There are two files that are changed:
A HTML file, to add a button that displays the solution, and a hidden textarea that has the solution to the exercise.
A JavaScript file, that gets the solution from the html file. And adds the event to the solution button to display the solution on the screen.

### DIFF
--- a/html/forms/tasks/form-structure/form-structure1.html
+++ b/html/forms/tasks/form-structure/form-structure1.html
@@ -53,25 +53,25 @@
 
     
     <textarea id="solution-code" style="display: none;">
-      <form>
-        <fieldset>
-          <legend>Personal details</legend>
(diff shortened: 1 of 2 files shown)


bsmth:
Thanks for the suggestion. In https://github.com/mdn/content/pull/40353, the solution was added to the page in a details element. You can see it here: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Forms_and_buttons#forms_and_buttons_5

I'm going to close for now. @chrisdavidmills does that make sense?

PR #788, opened 2025-03-11
Make the p text-color consistent with the provided design

youssef-el-atmani:
When working on that test **using this code** instead of working on it on the test home (mdn learn page), the end result looks slightly different from how it should look like.

Here is how it should look like: 
> This screen shot from the [test](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills:_HTML_text_basics#task_4) page

Screenshot from 2025-03-11 10-10-14

And this is how the end result looks like when using the original code provided:
> Look at the code at [codePen](https://codepen.io/youssef-elatmani/pen/vEYeXqR)
> Note that I swapped *music* with *books*, in the bellow preview. That only applied to me.

Screenshot from 2025-03-11 10-13-13

### DIFF
--- a/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
+++ b/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
@@ -26,7 +26,7 @@
         color: purple;
       }
 
-      p, em {
+      em {
         color: gray;
         margin: 0.5em 0;
       }
--- a/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
+++ b/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
@@ -24,9 +24,7 @@
       p, em {
         margin: 0.5em 0;
       }
-      p {
-        color: black;
-      }
+
       em {
         color: gray;
       }
--- a/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
+++ b/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
@@ -27,11 +27,13 @@
       }
 
       p, em {
-        color: gray;
         margin: 0.5em 0;
       }
       p {
         color: black;
+      }
+      em {
+        color: gray;
       }
 
       * {
--- a/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
+++ b/html/introduction-to-html/tasks/basic-text/basic-text3-download.html
@@ -26,9 +26,12 @@
         color: purple;
       }
 
-      em {
+      p, em {
         color: gray;
         margin: 0.5em 0;
+      }
+      p {
+        color: black;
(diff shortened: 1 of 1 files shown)


youssef-el-atmani:
The elements are black by default, so maybe the `property` that sets the `paragraph` to be black could be removed and will make no difference.

I don't know which one is better, so I'll keep that decision to whoever review this PR!

I would like to know if it is good idea to rely on the default style, or it is better to explicitly style the elements.

youssef-el-atmani:
Hello @chrisdavidmills, would you revise this PR please.

chrisdavidmills:
Hi @youssef-el-atmani! Of course, I'd be happy to review this.

Can you tell me which MDN page this example file is referenced from?

dipikabh:
Hey @chrisdavidmills, I too just started looking at this.
They're looking at the task here https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills:_HTML_text_basics#task_4 with the corresponding files at https://github.com/mdn/learning-area/tree/main/html/introduction-to-html/tasks/basic-text.

html/introduction-to-html/tasks/basic-text/basic-text3-download.html
```diff
@@ -27,9 +27,14 @@
       }
 
       p, em {
-        color: gray;
         margin: 0.5em 0;
       }
+      p {
```
chrisdavidmills, line 32: I'd remove this rule; there is no point setting the paragraph color to black, as that is the browser default anyway.

chrisdavidmills (changes requested):
@youssef-el-atmani looks better, but I would remove the default text coloring rule, as commented. I would also do the following:

1. Remove the h2 rule, as there is no `<h2>` in the example:
```
h2 {
  font-size: 1.6rem;
  margin: 0;
}
```

2. Update the selector list in the following rule so it is just `h1, strong`. Again, there is no `<h2>` to be selected.
```
h1, h2, strong {
  color: purple;
}
```

youssef-el-atmani:
Thank you @chrisdavidmills very much,
As you suggest I've **removed** the:
- unused `h2` declaration, 
- and the unnecessary `p` declaration.

chrisdavidmills (approved):
@youssef-el-atmani Looking great! Thanks so much for your MDN contribution!

Let's get this merged.

chrisdavidmills:
Ooop, we've still got @dipikabh as a pending reviewer. Did you have any further thoughts, Dipika? Feel free to approve and merge if you are happy with this.

dipikabh (approved):
Oops, didn't intend to block this. Thanks for reviewing, @chrisdavidmills.

PR #791, opened 2025-03-26
Fix out of bounds, stationary balls

olivia-evans-tulsa:
bugs solved:
x possible to be 0, meaning ball is stuck on x axis
y possible to be 0, meaning ball is stuck on y axis 
possible for velX and velY to be 0, meaning no movement


jeevanroy1434 (approved):

pepelsbey (approved):
Good catch! Thank you :)

PR #795, opened 2025-04-29
Add title attributes to match task

samlonginfo:
## What?

I've added title attributes to the two abbreviation tags (one for HTML and one for CSS), referencing their expansions.

## Why?

These changes reflect the requirements of the task found [here](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Test_your_skills/Advanced_HTML_text).


pepelsbey:
Hey! This is a sample file for anyone to use as a starting point in testing their skills. Feel free to practice in your fork :)

PR #812, opened 2025-07-22
Update index.html

SahilManav:
Updated paragraph text in getting-started/index.html for better clarity on basic HTML structure.

### DIFF
--- a/html/introduction-to-html/getting-started/index.html
+++ b/html/introduction-to-html/getting-started/index.html
@@ -6,6 +6,6 @@
     <title>My test page</title>
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


dipikabh:
Hi @SahilManav, welcome to MDN 🎉 

We'll be keeping the code as-is to match the examples we've used in the articles in the Learn area (see [Anatomy of an HTML document](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Structuring_content/Basic_HTML_syntax#anatomy_of_an_html_document) for example).

Hope you'll continue contributing to MDN.

PR #845, opened 2026-07-02

yousukka:


dipikabh (approved):
Thanks for sending the fix!

oto-code:
The error on line 87 was intentional and is a part of https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/What_went_wrong

PR #848, opened 2026-08-07
Remove unreferenced examples from accessibility and css directories

chrisdavidmills:
This PR is part of a recent project to clean up the [Learn](https://developer.mozilla.org/en-US/docs/Learn_web_development)-related repos by removing all files that are no longer referenced in [content](https://github.com/mdn/content), hopefully saving us a lot of maintenance work in the future.

This PR focuses just on the accessibility and css directories; it has gotten big enough already. I've removed all the files that are no longer needed here, due to them being covered by live examples in the content, or [shared-assets](https://github.com/mdn/shared-assets/) files, or by simply not being used anymore at all.

Note that I have made some changes to the content repo too, to change some things around and move some examples into the pages as live samples, thereby allowing me to remove more items from this repo: see https://github.com/mdn/content/pull/45046.

These two PRs need to be merged simultaneously to avoid breakage.


Rumyra (approved):
I've crossed checked against the other pr and all looks good

PR #853, opened 2026-09-21
Remove unneeded line

chrisdavidmills:
In https://github.com/mdn/content/issues/45793, the OP remarks how the instructions tell the reader to remove `margin: 0 auto;` from the `.infobox`, but the code shows it still there.

I've tested it, and it is definitely not needed. This PR removed it from the corresponding live example too.


pepelsbey (approved):
Good catch! Thank you 🙂

